PR #1, opened 2025-03-01
index json

mgorny:
Here's a prototype implementation. Given a `directory`, it reads variant metadata from all `*.whl` files, and writes a `variants.json` file.

Here's an example file from `mockhouse/pep-xxx-variants/dummy-project/`, pretty-printed using `json_pp` (the output is compacted):

```json
{
   "provider-requires" : [
      "provider_fictional_hw",
      "provider_fictional_tech"
   ],
   "variants" : {
      "36266d4d" : {
         "fictional_hw" : {
            "architecture" : "HAL9000",
            "compute_accuracy" : "0",
            "compute_capability" : "6",
            "humor" : "2"
         }
      },
      "4f8ae729" : {
         "fictional_hw" : {
            "architecture" : "tars",
            "compute_accuracy" : "8",
            "compute_capability" : "8",
            "humor" : "10"
         }
      },
      "57768a46" : {
         "fictional_tech" : {
            "quantum" : "FOAM",
            "risk_exposure" : "1000000000",
            "technology" : "improb_drive"
         }
      },
      "6b4c8391" : {
         "fictional_hw" : {
            "architecture" : "deepthought",
            "compute_accuracy" : "10",
            "compute_capability" : "10",
            "humor" : "0"
         },
         "fictional_tech" : {
            "quantum" : "FOAM"
         }
      },
      "9091cdc4" : {
         "fictional_tech" : {
            "quantum" : "SUPERPOSITION",
            "risk_exposure" : "25",
            "technology" : "auto_chef"
         }
      },
      "e684be6f" : {
         "fictional_hw" : {
            "architecture" : "mother",
            "compute_capability" : "4"
         }
      }
   }
}
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ variantlib = "variantlib.commands.main:main"
 [project.entry-points."variantlib.actions"]
 analyze_wheel = "variantlib.commands.analyze_wheel:analyze_wheel"
 analyze_platform = "variantlib.commands.analyze_platform:analyze_platform"
+generate_index_json = "variantlib.commands.generate_index_json:generate_index_json"
 
 [tool.pytest.ini_options]
 testpaths = ["tests/"]
--- a/variantlib/commands/generate_index_json.py
+++ b/variantlib/commands/generate_index_json.py
@@ -1,0 +1,87 @@
+import argparse
+import email.parser
+import email.policy
+import json
+import logging
+import pathlib
+import zipfile
+
+from variantlib.meta import VariantMeta
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def generate_index_json(args):
+    parser = argparse.ArgumentParser(
+        prog="generate_index_json",
+        description="Generate a JSON index of all package variants",
+    )
+    parser.add_argument(
+        "-d",
+        "--directory",
+        type=pathlib.Path,
+        required=True,
+        help="Directory to process",
+    )
+
+    parsed_args = parser.parse_args(args)
+
+    directory = parsed_args.directory
+
+    if not directory.exists():
+        raise FileNotFoundError(f"Directory not found: `{directory}`")
+    if not directory.is_dir():
+        raise NotADirectoryError(f"Directory not found: `{directory}`")
+
+    metadata_parser = email.parser.BytesParser(policy=email.policy.compat32)
+    known_variants = {}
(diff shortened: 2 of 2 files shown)


DEKHTIARJonathan:
LGTM - Merged

DEKHTIARJonathan:
@mgorny I merged because I want to leave you full engineering freedom on your idea :) 

I have one question: why "checking on the plugins" ? I think it's a big big problem. I don't see why it's needed - we are just parsing wheel variants one by one and aggregating/formatting in json.

 I don't think we need to validate that the plugin exists/is known. What if the plugin is privately published / not on PyPI (like on PyT index).
 
 From my perspective it's a "descriptive process" not a guarantee that the plugin exists - nor where the plugin comes from.
 In the future - we might add a `pinned version of the plugin and index URL` almost like `build-dependencies` let's think about it and keep it simple for now 👍 
 
 Let's prove the concept from a high level perspective - convince ourselves we are in a really solid engineering path.

mgorny:
> I have one question: why "checking on the plugins" ? I think it's a problem. I don't see why it's needed - we are just parsing wheel variants one by one and aggregating/formatting in json.

The way I see it, the resolver needs to know which plugins to install, in order to be able to use variants. In other words, the way I see it:

1. `pip` fetches `variants.json`.
2. `pip` installs plugins from `provider-requires`.
3. Now `pip` can query the plugins to select variants.

Otherwise, we would require the users to know up front which plugins are used by every package they install (including their indirect dependencies), or they would always hit a fallback to the "non-variant" wheel.

Say, `torch` has variants for different CUDA versions. By default, if you create a new virtual environment and `pip install torch`, it won't be able to match CUDA variants because there's no CUDA variant provider. However, if it queries `variants.json` first, it can install the necessary provider and select the optimal variant.

DEKHTIARJonathan:
I don't disagree with that statement but why is it called in the part that generate the `json` - that part will be run by the index most of the time who doesn't need any plugin.

mgorny:
Ah, it was just a convenient place to put it in for the demo.

DEKHTIARJonathan:
OIC - that part of the code +/- already exists. Check here: https://github.com/wheelnext/variantlib/blob/main/variantlib/platform.py

mgorny:
Yes, that's where I adapted the function from. Didn't want to factor out the common bits for now, while it's all being drafted still.

PR #4, opened 2025-03-20
Relax whitespace requirements in `VariantMeta.from_str()`

mgorny:
Allow any (or no) whitespace between `::` separators in `VariantMeta.from_str()`, in order to make it more feasible for parsing user input.  Unless I'm missing something, there should be no need to strictly ensure that input is strictly formatted as `A :: B :: C` (rather than, say, `A::B::C`), as long as we properly format output. This can be especially useful for command-line arguments where spaces would have to be quoted.

### DIFF
--- a/tests/test_meta.py
+++ b/tests/test_meta.py
@@ -90,9 +90,16 @@ def test_failing_regex_value():
             _ = VariantMeta(provider="provider", key="key", value=f"val{c}ue")
 
 
-def test_from_str_valid():
+@pytest.mark.parametrize(
+    "input_str",
+    [
+        "OmniCorp :: access_key :: secret_value",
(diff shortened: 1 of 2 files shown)


DEKHTIARJonathan (changes requested):
That is true indeed - it's not a `hard requirement per say`.

However, this is mostly intended to input str from the `METADATA` file which will only be formatted with the ` :: ` sequence (for consistency).
So I'm not entirely sure of the value of this change - I don't see why/where it's gonna be useful.

In any case if you want to proceed with this, do you mind adding a few unittests checking that the objects created are identical ?

mgorny:
Added a few variants to `test_from_str_valid`.

DEKHTIARJonathan (approved):

PR #5, opened 2025-03-21
Abstract away plugin lookup and querying

mgorny:
Add a `PluginLoader` class that handles loading and querying plugins.
This removes some duplicate code, prepares for plugins having more
than one method in the future, and enables using a single cached plugin
list, rather than repeatedly querying and instantiating them.


mgorny:
(I've switched to a static method)

DEKHTIARJonathan:
https://github.com/mgorny/variantlib/pull/1

mgorny:
Resolved merge conflicts after `attrs` removal.

PR #8, opened 2025-03-24
plugins: Rename `run()` to `get_supported_configs()`, add tests

mgorny:
Rename the `run()` API to `get_supported_configs()`, to prepare for adding other plugin methods.  Fix bugs introduced while introducing the singleton pattern.  Add tests.


DEKHTIARJonathan (approved):
LGTM

PR #10, opened 2025-03-25
Add a `Protocol` and an ABC for plugin classes

mgorny:
This uses a twofold approach:

The `PluginType` provides an `Interface` defining how a plugin class looks like, "duck typing" style. `providerlib` uses it to verify that the class returned by the entry point matches the prototype (i.e. defines all the methods that are necessary), so that we can tell immediately "the entry point returned a class that doesn't implement everything we need", rather than the code failing sometime later over calling a nonexistent method.

The `PluginBase` provides an abstract base class that could be used to implement plugins. It's not obligatory (i.e. a class that doesn't inherit from `PluginBase` will still satisfy the interface), but it helps people by making their plugins explode immediately if they don't implement some obligatory method. Over time, we can also use them to provide a default implementation for optional methods.

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -3,11 +3,12 @@ from typing import Any, Optional
 
 import pytest
 
-from variantlib.config import KeyConfig,ProviderConfig
+from variantlib.base import PluginBase
+from variantlib.config import KeyConfig, ProviderConfig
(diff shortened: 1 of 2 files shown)


DEKHTIARJonathan (approved):
LGTM

PR #12, opened 2025-03-25
Add an API to describe variant using labels

mgorny:
Add a minimal API that lets plugins process the variant description and add short labels that could be used to describe the wheel. The labels are entirely optional -- plugins may not return any, and clients may ignore them.  The design also assumes that the client is responsible for choosing how many labels to use before truncating -- though I suppose I'll add a helper function to `variantlib` for that purpose.

The design assumes that plugin get complete unfiltered `VariantDescription` -- and therefore also see variant metadata from other plugins.  I don't think that's really a problem, though it assumes that the plugin must compare both against provider and key names.

At this point, the protocol makes the API mandatory, even if it would only return an empty list unconditionally.  Perhaps we should make the function optional somehow instead.

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -61,6 +61,13 @@ class MockedPluginC(PluginBase):
         return ret
 
 
+class MockedPluginD:
+    namespace = "plugin_without_labels"
+
+    def get_supported_configs(self) -> Optional[ProviderConfig]:
+        return None
+
+
 class ClashingPlugin(PluginBase):
     namespace = "test_plugin"
 
@@ -112,6 +119,11 @@ def mocked_plugin_loader(session_mocker):
             name="incompatible_plugin",
             value="tests.test_plugins:MockedPluginC",
             plugin=MockedPluginC,
+        ),
+        MockedEntryPoint(
+            name="no_labels",
+            value="tests.test_plugins:MockedPluginD",
+            plugin=MockedPluginD,
         ),
     ]
     yield PluginLoader()
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -5,6 +5,7 @@ import pytest
 
 from variantlib.base import PluginBase
 from variantlib.config import KeyConfig, ProviderConfig
+from variantlib.meta import VariantDescription, VariantMeta
 from variantlib.plugins import PluginLoader
 
 
@@ -20,6 +21,12 @@ class MockedPluginA(PluginBase):
             ],
         )
 
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        for meta in variant_desc:
+            if meta.namespace == self.namespace and meta.key == "key1":
+                return [meta.value.removeprefix("val")]
+        return []
+
 
 # NB: this plugin deliberately does not inherit from PluginBase
 # to test that we don't rely on that inheritance
@@ -34,12 +41,24 @@ class MockedPluginB:
             ],
         )
 
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        if VariantMeta(self.namespace, "key3", "val3a") in variant_desc:
+            return ["sec"]
+        return []
+
 
 class MockedPluginC(PluginBase):
-    namespace = "incompatible_plugin"
+    namespace = "other_plugin"
 
     def get_supported_configs(self) -> Optional[ProviderConfig]:
         return None
+
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        ret = []
+        for meta in variant_desc:
+            if meta.namespace == self.namespace and meta.value == "on":
+                ret.append(meta.key)
+        return ret
 
 
 class ClashingPlugin(PluginBase):
@@ -47,6 +66,13 @@ class ClashingPlugin(PluginBase):
 
     def get_supported_configs(self) -> Optional[ProviderConfig]:
         return None
+
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        ret = []
+        for meta in variant_desc:
+            if meta.namespace == self.namespace and meta.value == "on":
+                ret.append(meta.key)
+        return ret
 
 
 @dataclass
@@ -137,3 +163,36 @@ def test_namespace_clash(mocker):
     assert "same namespace test_plugin" in str(exc)
     assert "test-plugin" in str(exc)
     assert "clashing-plugin" in str(exc)
+
+
+@pytest.mark.parametrize("variant_desc,expected",
+[
+    (VariantDescription([
+        VariantMeta("test_plugin", "key1", "val1a"),
+        VariantMeta("test_plugin", "key2", "val2b"),
+        VariantMeta("second_plugin", "key3", "val3a"),
+        VariantMeta("other_plugin", "flag2", "on"),
+    ]), ["1a", "sec", "flag2"]),
+    (VariantDescription([
+        # note that VariantMetas don't actually have to be supported
+        # by the system in question -- we could be cross-building
+        # for another system
+        VariantMeta("test_plugin", "key1", "val1f"),
+        VariantMeta("test_plugin", "key2", "val2b"),
+        VariantMeta("second_plugin", "key3", "val3a"),
+    ]), ["1f", "sec"]),
+    (VariantDescription([
+        VariantMeta("test_plugin", "key2", "val2b"),
+        VariantMeta("second_plugin", "key3", "val3a"),
+    ]), ["sec"]),
+    (VariantDescription([
+        VariantMeta("test_plugin", "key2", "val2b"),
+    ]), []),
+    (VariantDescription([
+        VariantMeta("test_plugin", "key2", "val2b"),
+        VariantMeta("other_plugin", "flag1", "on"),
+        VariantMeta("other_plugin", "flag2", "on"),
+    ]), ["flag1", "flag2"]),
+])
+def test_get_variant_labels(mocked_plugin_loader, variant_desc, expected):
+    assert mocked_plugin_loader.get_variant_labels(variant_desc) == expected
--- a/variantlib/base.py
+++ b/variantlib/base.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod, abstractproperty
 from typing import Protocol, runtime_checkable
 
 from variantlib.config import ProviderConfig
+from variantlib.meta import VariantDescription
 
 
 @runtime_checkable
@@ -17,6 +18,10 @@ class PluginType(Protocol):
         """Get supported configs for the current system"""
         ...
 
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        """Get list of short labels to describe the variant"""
+        ...
+
 
 class PluginBase(ABC):
     """An abstract base class that can be used to implement plugins"""
@@ -26,3 +31,6 @@ class PluginBase(ABC):
 
     @abstractmethod
     def get_supported_configs(self) -> ProviderConfig: ...
+
+    def get_variant_labels(self, variant_desc: VariantDescription) -> list[str]:
+        return []
--- a/variantlib/base.py
+++ b/variantlib/base.py
@@ -5,7 +5,6 @@ from variantlib.config import ProviderConfig
 from variantlib.meta import VariantDescription
 
 
-@runtime_checkable
 class PluginType(Protocol):
     """A protocol for plugin classes"""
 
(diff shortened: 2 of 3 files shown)


mgorny:
Marking this as draft until I've got meson-python counterpart ready.

DEKHTIARJonathan:
@mgorny I read this PR (because I was trying to rebase it) and I think I disagree with what is being done here (as an objective - not how it's done).

I don't think this is a good idea.I think that should be a package maintainer API:

`python -m build --variant_alias=cuda12_sm90`

Given that these names have no way to guarantee unicity (unlike the hash). We shouldn't pretend they are "part of the variant API". They are not. They are convenience identifiers - mostly for maintainers / power users. They are not part of either the variant resolution nor checked in any way for unicity / validity (beyond a regex and max length maybe).

Plus different packages will want to "name the same variant" differently because they will focus on different aspects.

And lastly I don't think the alias should include every "features" inside the alias: `+cuda12+sm90+avx512+avx2+sse+...`. This is exactly what I'm trying to prevent.

mgorny:
@rgommers, what do you think?

mgorny:
> 1. I'm a little surprised that `variantlib` would be involved in this at all. It seems to me like it should be an interface between each plugin and the build backend. That's how I sketched it in https://github.com/wheelnext/pep_xxx_wheel_variants/pull/9/files. Any thoughts on the tradeoffs here?

My idea is that `variantlib` is a convenient place to provide plugin lookup code and interface, so we don't end up independently searching for plugins all over the place. As you can see, the API here is pretty much limited to querying labels from the plugins, and passing them back.

rgommers:
> My idea is that `variantlib` is a convenient place to provide plugin lookup code and interface, so we don't end up independently searching for plugins all over the place.

Yeah I can see the case for that. It can/will work either way I think. And whatever the design is, it needs a documented and stable interface anyway, because `variantlib` should be reimplementable - so a build backend anyway has the choice to query plugins directly if it really wants to for some reason.

DEKHTIARJonathan:
> I was indeed expecting variantlib to only deal with the hashes.

I really think this is the best idea. Hence I'm pushing back here. 

In my perspective, we have to be careful with adding this "variant alias" feature. It's a non-essential feature though very nice to have. Modifying the filename (to include `vhash`) is already something that people might get defensive about. I really want to keep the "frustration on the low side". 

> I'm a little surprised that variantlib would be involved in this at all. It seems to me like it should be an interface between each plugin and the build backend.

I agree here.

> They should be free to override it of course, but forcing non-uniformity by design seems worse than having good defaults. In addition, the UX is worse without defaults, because you have to add something like --variant_by_alias=cuda12_sm90 to every build invocation.

I think here the "disconnect" is more on the "usecase/scenario" and we should answer that question first.
- Why do we want that feature ? To answer what kind of workflow ?
- What type of users are we expecting to use `variant_aliases` ?

I think it's reasonable to say >95-98% of users will never see them - never be aware of them - never even understand them in most cases. I would argue if that statement is not correct, we need to re-engineer the "variant resolution algorithm" because that's not good. 

If the idea is to give a "unique name" to each [major ?] variant to allow people to fix "variant resolution issue" then I would argue that there is no need for correlation for feature / name. A high level name that can be documented is good enough. And that most importantly will be short because it's an alias (e.g. `cu128_hopper`, `skylake_ompi`, etc.)

Now if the idea is to provide package maintainers a way to in local identify different variants to the "smallest degree of granularity", this is a completely use-case and they can be flagged as "development wheels" and not be published "as-is".

> And since in CI you're often building in a matrix, it will get even worse with string interpolation (e.g., --variant_by_alias=${{ matrix.buildplat[4] }}_${{ matrix.buildplat[5] }}.

Depending on the "usecase" I am not convinced every variants need an "alias".

**Question:** so now that clearly opens the door to alias conflicts. Because obviously I can have  `matrix.buildplat[1]`, `matrix.buildplat[2]`, `matrix.buildplat[3]` varying and generating the same alias. It kinda reinforces that not every variants should have an alias. Only those we do want to flag as "major variants" and document.

DEKHTIARJonathan:
> I was indeed expecting variantlib to only deal with the hashes.

> I'm a little surprised that variantlib would be involved in this at all. It seems to me like it should be an interface between each plugin and the build backend.

That was the original idea, though I do agree with @mgorny here that we should increase the scope. It's much cleaner to provide a very strong, well tested complete implementation of the variant standard behavior. Rather than 50 projects implementing 15% of the complete standard.

It's both a reference implementation and a support library for the entire ecosystem of tools that will build on top of wheel variant features.

> They should be free to override it of course, but forcing non-uniformity by design seems worse than having good defaults. In addition, the UX is worse without defaults, because you have to add something like --variant_by_alias=cuda12_sm90 to every build invocation.

I think here the "disconnect" is more on the "usecase/scenario" and we should answer that question first.
- Why do we want that feature ? To answer what kind of workflow ?
- What type of users are we expecting to use `variant_aliases` ?

I think it's reasonable to say >95-98% of users will never see them - never be aware of them - never even understand them in most cases. I would argue if that statement is not correct, we need to re-engineer the "variant resolution algorithm" because that's not good. 

If the idea is to give a "unique name" to each [major ?] variant to allow people to fix "variant resolution issue" then I would argue that there is no need for correlation for feature / name. A high level name that can be documented is good enough. And that most importantly will be short because it's an alias (e.g. `cu128_hopper`, `skylake_ompi`, etc.)

Now if the idea is to provide package maintainers a way to in local identify different variants to the "smallest degree of granularity", this is a completely use-case and they can be flagged as "development wheels" and not be published "as-is".

> And since in CI you're often building in a matrix, it will get even worse with string interpolation (e.g., --variant_by_alias=${{ matrix.buildplat[4] }}_${{ matrix.buildplat[5] }}.

Depending on the "usecase" I am not convinced every variants need an "alias".

**Question:** so now that clearly opens the door to alias conflicts. Because obviously I can have  `matrix.buildplat[1]`, `matrix.buildplat[2]`, `matrix.buildplat[3]` varying and generating the same alias. It kinda reinforces that not every variants should have an alias. Only those we do want to flag as "major variants" and document.

> That seems a little far-fetched since we don't want that many variables to begin with.

I don't disagree. But unless you block it people will do it. My whole point is "if we consider a behavior undesirable - let's close the door to that behavior". I don't want to limit the number of features a variant can declare (arbitrary philosophy), but then consequently I need to block that feature to prevent the filenames to become arbitrarily long.

--------------------------------

My overall perspective on this variant alias feature is that we have to be careful.

Adding this "variant alias" feature,  a non-essential feature though very nice to have, implies to modify the filename a lot more than absolutely necessary (`vhash` is the only essential part). Any modification of that filename is already something that people might get defensive about. I really want to keep the "frustration on the low side". I do believe people will hear us out that we need to add the hash to the filename for so many good reasons. I am not convinced at all the community will let us add anything else beyond that hash, and almost certainly not an arbitrary long string. Especially that this alias comes with no unicity guarantee. We should totally include it in the PEP, and ask for the best case scenario. However if we are not reasonable in our approach we will make an easy target (and we are already an easy one).

DEKHTIARJonathan:
> I was indeed expecting variantlib to only deal with the hashes.

> I'm a little surprised that variantlib would be involved in this at all. It seems to me like it should be an interface between each plugin and the build backend.

That was the original idea, though I do agree with @mgorny here that we should increase the scope. It's much cleaner to provide a very strong, well tested complete implementation of the variant standard behavior. Rather than 50 projects implementing 15% of the complete standard.

It's both a reference implementation and a support library for the entire ecosystem of tools that will build on top of wheel variant features.

> They should be free to override it of course, but forcing non-uniformity by design seems worse than having good defaults. In addition, the UX is worse without defaults, because you have to add something like --variant_by_alias=cuda12_sm90 to every build invocation.

I think here the "disconnect" is more on the "usecase/scenario" and we should answer that question first.
- Why do we want that feature ? To answer what kind of workflow ?
- What type of users are we expecting to use `variant_aliases` ?

I think it's reasonable to say >95-98% of users will never see them - never be aware of them - never even understand them in most cases. I would argue if that statement is not correct, we need to re-engineer the "variant resolution algorithm" because that's not good. 

If the idea is to give a "unique name" to each [major ?] variant to allow people to fix "variant resolution issue" then I would argue that there is no need for correlation for feature / name. A high level name that can be documented is good enough. And that most importantly will be short because it's an alias (e.g. `cu128_hopper`, `skylake_ompi`, etc.)

Now if the idea is to provide package maintainers a way to in local identify different variants to the "smallest degree of granularity", this is a completely use-case and they can be flagged as "development wheels" and not be published "as-is".

> And since in CI you're often building in a matrix, it will get even worse with string interpolation (e.g., --variant_by_alias=${{ matrix.buildplat[4] }}_${{ matrix.buildplat[5] }}.

Depending on the "usecase" I am not convinced every variants need an "alias".

**Question:** so now that clearly opens the door to alias conflicts. Because obviously I can have  `matrix.buildplat[1]`, `matrix.buildplat[2]`, `matrix.buildplat[3]` varying and generating the same alias. It kinda reinforces that not every variants should have an alias. Only those we do want to flag as "major variants" and document.

> That seems a little far-fetched since we don't want that many variables to begin with.

I don't disagree. But unless you block it people will do it. My whole point is "if we consider a behavior undesirable - let's close the door to that behavior". I don't want to limit the number of features a variant can declare (arbitrary philosophy), but then consequently I need to block that feature to prevent the filenames to become arbitrarily long.

--------------------------------

My overall perspective on this variant alias feature is that we have to be careful.

Adding this "variant alias" feature,  a non-essential feature though very nice to have, implies to modify the filename a lot more than absolutely necessary (`vhash` is the only essential part). Any modification of that filename is already something that people might get defensive about. I really want to keep the "frustration on the low side". I do believe people will hear us out that we need to add the hash to the filename for so many good reasons. I am not convinced at all the community will let us add anything else beyond that hash, and almost certainly not an arbitrary long string. Especially that this alias comes with no unicity guarantee. We should totally include it in the PEP, and ask for the best case scenario. However if we are not reasonable in our approach we will make an easy target (and we are already an easy one). 

I do not care much about how this is implemented. I don't care much if it's fully manual or auto + override. But we need a very very clear use-case, and be able to explain why we believe it's really important to have this. Because arguably - if the variant resolution algorithm is good - then most users will never see these aliases.

- If we target users consuming from an index, this is a totally functional solution:
Screenshot 2025-03-31 at 15 59 09

- If we target users using wheels locally (like @rgommers you were mentioning) - then we can remove any limit at all and just allow this scheme locally/from disk.

mgorny:
I'd like to merge #20 finish, but then I'd like to revisit this, so I'd appreciate some guidance on what's your preferred course of action here. FWICS, the main options here are:

1. rebase/update it as-is, i.e. with labels being added automatically
2. abandon variantlib part, and instead add a meson-python option to add labels explicitly

DEKHTIARJonathan:
There is absolutely no consensus among us about `label`. I would suggest to hold off entirely on the feature.
This feature has a high potential to become an easy target to attack the PEP - @warsaw was suggesting to even add that behavior in a second PEP to avoid unnecessary blockers to the greater objective.

I'm sympathetic to the objective being achieved here, but I'm not convinced at all:
1. It will accepted at all in the PEP review process - I'm actually quite pessimistic about it.
2. I don't want this to become a point of contention and create unnecessary fuss on DPO.

Before we make a decision whether or not to add `variant alias` and in what form to the initial PEP - I'd really like that we agree at least among all core WheelNext members that we have picked the best "initial idea" and we are all comfortable to support this idea. Right now I don't think we have such a consensus.

DEKHTIARJonathan:
There is absolutely no consensus among us about `label`. I would suggest to hold off entirely on the feature.
This feature has a high potential to become an easy target to attack the PEP - @warsaw was suggesting to even add that behavior in a second PEP to avoid unnecessary blockers to the greater objective.

I'm sympathetic to the objective being achieved here, but I'm not convinced at all:
1. It will accepted at all in the PEP review process - I'm actually quite pessimistic about it.
2. I don't want this to become a point of contention and create unnecessary fuss on DPO.

Before we make a decision whether or not to add `variant alias` and in what form to the initial PEP - I'd really like that we agree at least among all core WheelNext members that we have picked the best "initial idea" and we are all comfortable to support this idea. Right now I don't think we have such a consensus.

@mgorny @rgommers @msarahan @warsawnv @emmatyping-nv @aterrel @charliermarsh @konstin @seemethere @atalman

charliermarsh:
Thanks for the summary @DEKHTIARJonathan. Personally, I agree that this feels non-critical (and that we have larger points to align on) -- I would also prefer to defer this for now.

warsaw:
I know @rgommers feels strongly that aliases should be included in the first (and only) PEP, but you also know that I agree that I think this will be a bone of contention that distracts from the core bits of the PEP.   I think it would be perfectly fine to lay out the argument for aliases in an Open Questions section and explicitly state that a) it's not critical to the functional success of variants, b) enhances the usability of variants, c) may be proposed in a companion PEP.  If we're wrong about that, then this will come up in the DPO thread and we can update the variants spec to include aliases pretty easily.

DEKHTIARJonathan:
Closing in favor of #108 . If this is needed - feel free to re-open

PR #13, opened 2025-03-26
Fix `logger.warning()` arguments

mgorny:


variantlib/combination.py
```diff
@@ -102,15 +102,14 @@ def variant_sort_key_gen(desc: VariantDescription) -> Generator[tuple]:
...
+            {"provider": " ".join(missing_providers)},
         )
 
     for provider, provider_missing_keys in missing_keys.items():
         logger.warning(
             "The %(provider)s provider does not provide the following expected keys: "
             "%(missing_keys)s; some variants will be ignored",
-            provider=provider,
-            missing_keys=" ".join(provider_missing_keys),
+            {"provider": provider, "missing_keys": " ".join(provider_missing_keys)},
```
DEKHTIARJonathan, line 112: FYI this is expected - this informs that the platform is not compatible with feature X, Y and Z. This message could totally be confusing.

Example: if you're not compatible with `AVX512` - you would not report it at all.

PR #14, opened 2025-03-26
plugins: Add a public `name` property and use it as primary identifier

mgorny:
Add a public `name` property and use it to identify plugins over the current mix of entry point and `ProviderConfig` names. We require that `ProviderConfig` specifies the same name now. Related to https://github.com/wheelnext/pep_xxx_wheel_variants/issues/16.

I've also made the distribution name/version entirely optional, as entry points API apparently permits entry points without a distribution.

Do you want me to add a `version` field as well?

Also, I'm wondering if we shouldn't put our demo plugins in a single repository. Updating them for API changes separately is a bit of a hassle now.

### DIFF
--- a/tests/artifacts/expected.json
+++ b/tests/artifacts/expected.json
@@ -1,783 +1,783 @@
 [
     [
         {
-            "provider": "custom_hw",
+            "namespace": "custom_hw",
             "key": "driver_version",
             "value": "1.3"
         },
         {
-            "provider": "custom_hw",
+            "namespace": "custom_hw",
             "key": "hw_architecture",
             "value": "3.4"
         },
         {
-            "provider": "networking",
+            "namespace": "networking",
             "key": "speed",
             "value": "10GBPS"
         }
     ],
     [
         {
-            "provider": "custom_hw",
+            "namespace": "custom_hw",
             "key": "driver_version",
             "value": "1.3"
         },
         {
-            "provider": "custom_hw",
+            "namespace": "custom_hw",
             "key": "hw_architecture",
             "value": "3.4"
         },
         {
-            "provider": "networking",
+            "namespace": "networking",
             "key": "speed",
             "value": "1GBPS"
         }
     ],
     [
         {
-            "provider": "custom_hw",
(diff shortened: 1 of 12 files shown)


DEKHTIARJonathan (changes requested):
After numerous conversations - I think we should change something here.

Let's introduce the concept of `namespace`:

```python
ProviderConfig(
  namespace="nvidia",
  ...
)
```

and 

```python
class PluginType(Protocol):
    """A protocol for plugin classes"""

    @property
    def namespace(self) -> str:
        """Get provider namespace"""
        ...
```

Using the words `name` and `provider` was a poor choice (I made) initially. It suggests that these 3 things are linked - when really there are not:
- plugin package name => has no influence
- entrypoint name => has no influence
- provider name(space) => only meaningful attribute

`namespace` is the right wording - let's remove reference to any other wording to avoid confusion.

### side change

It's important to ensure `variantlib` errors when two plugins use the same `namespace`.

Expected behavior in case of namespace conflict:
1. log.warning() an explicit message explaining that package A and B are mutually exclusive. Remove one.
2. Disable variant behavior by returning `None` or `[ ]` (therefore no variant is being picked up)

mgorny:
> Disable variant behavior by returning None or [ ] (therefore no variant is being picked up)

Shouldn't we just proceed with the remaining plugins? Otherwise we're creating inconsistency with the situation where neither of the plugins were installed in the first place.

DEKHTIARJonathan:
> > Disable variant behavior by returning None or [ ] (therefore no variant is being picked up)
> 
> Shouldn't we just proceed with the remaining plugins? Otherwise we're creating inconsistency with the situation where neither of the plugins were installed in the first place.

No no... I'm against that. The best we can do is to create an explicit error to force the user to fix it. I could even be convinced to entirely have the `pip install package` command fail. It's far better user experience to have an explicit error telling you what the problem is and fixing it than trying to do a best effort recovery that we know is not going to be what the user expect.

DEKHTIARJonathan (approved):
LGTM

PR #15, opened 2025-03-26
[Refactoring] CI Pre-Commit added & Code Refactoring - Removing Singleton

DEKHTIARJonathan:
- A lot of code cleaning - Adding `pre-commit` to keep the codebase in a clean state.
- Removing the singleton `metaclass`
- Adding an `api.py` file to classify what are the most important APIs and what are "internal mechanics"


tests/test_plugins.py
```diff
@@ -1,17 +1,19 @@
...
+from variantlib.config import KeyConfig
+from variantlib.config import ProviderConfig
+from variantlib.loader import PluginLoader
 
 
 class MockedPluginA(PluginBase):
     namespace = "test_plugin"
 
-    def get_supported_configs(self) -> Optional[ProviderConfig]:
+    def get_supported_configs(self) -> Optional[ProviderConfig]:  # noqa: UP007
```
mgorny, line 16: I'm going to try and see if we can set a lower Python target instead. After all, the problem is that it insists on upgrading to a newer version, breaking compatibility.

mgorny: Doesn't ruff effectively supersede pylint at this point?

mgorny: Or perhaps just disable `UP*` checks in pylint — to the best of my knowledge, the pyupgrade part is well covered in ruff.

DEKHTIARJonathan: Pylance and RUFF don't cover the same features.

One doesn't replace the other though they overlap in the type checking functionality.

Maybe pylance type checking could be deactivated. Need to check

mgorny:
I'll merge it now to avoid things diverging more, and we'll see if we can improve it afterwards.

PR #17, opened 2025-03-27
Support using custom implementation of key configs in API

mgorny:
Add support for plugins implementing their own key config type rather than using the one from `variantlib`.  To account for that, introduce a `KeyConfigType` protocol that verifies that all mandatory fields are present, and convert the data passed by plugins to `KeyConfig` in plugin loader.

This is a step towards detaching the specification from `variantlib` internals.  Allowing an arbitrary implementation should improve compatibility with non-Python implementations, and also avoid potential issues when plugins use a different version of `variantlib` than the one vendored in `pip`.

See also: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/11

### DIFF
--- a/variantlib/loader.py
+++ b/variantlib/loader.py
@@ -20,8 +20,8 @@ class classproperty(property):  # noqa: N801
 class PluginLoader:
     """Load and query plugins"""
 
-    _plugins = {}
-    _dist_names = {}
+    _plugins: dict[str, PluginType] = {}
+    _dist_names: dict[str, str] = {}
 
     def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


variantlib/loader.py
```diff
@@ -57,7 +59,7 @@ def load_plugins(cls) -> None:
                 # Instantiate the plugin
                 plugin_instance = plugin_class()
                 assert isinstance(plugin_instance, PluginType)
-            except Exception:  # noqa: PERF203
```
mgorny, line 60: FYI, pre-commit did this.

DEKHTIARJonathan (approved):
LGTM

PR #20, opened 2025-03-29
Add API to get all valid configs and to validate variants

mgorny:
Add some API for builders, in particular:

- to get the list of all valid configs (e.g. for UI purposes)
- to validate that the passed variant description is valid (e.g. to reject building variants that don't match plugins)


variantlib/loader.py
```diff
@@ -127,6 +127,49 @@ def get_supported_configs(cls) -> dict[str, ProviderConfig]:
...
+            if not key_configs:
+                logging.error(
+                    "Provider: %(namespace)s does not support any configs!",
+                    {"namespace": namespace},
+                )
+                continue
+
+            for key_cfg in key_configs:
+                if not isinstance(key_cfg, KeyConfigType):
+                    logging.error(
```
DEKHTIARJonathan, line 156: Please change to a complete exception raise. No "log and ignore" inside `variantlib`.
If there is a single error or anything unexpected: Exception => Variant behavior deactivated

variantlib/loader.py
```diff
@@ -127,6 +127,49 @@ def get_supported_configs(cls) -> dict[str, ProviderConfig]:
...
+            if not isinstance(key_configs, list):
+                logging.error(
+                    "Provider: %(namespace)s returned an unexpected type: "
+                    "%(type)s - Expected: `list[KeyConfig]`. Ignoring...",
+                    {"namespace": namespace, "type": type(key_configs)},
+                )
+                continue
+
+            if not key_configs:
+                logging.error(
```
DEKHTIARJonathan, line 148: Same raise issue and stop

variantlib/loader.py
```diff
@@ -127,6 +127,49 @@ def get_supported_configs(cls) -> dict[str, ProviderConfig]:
...
+    @VariantCache()
+    def get_all_configs(cls) -> dict[str, ProviderConfig]:
+        """Get a mapping of namespaces to all valid configs"""
+
+        provider_cfgs = {}
+        for namespace, plugin_instance in cls._plugins.items():
+            key_configs = plugin_instance.get_all_configs()
+
+            if not isinstance(key_configs, list):
+                logging.error(
```
DEKHTIARJonathan, line 140: raise issue and stop

DEKHTIARJonathan (changes requested):

DEKHTIARJonathan:
@mgorny : we need to be more careful to follow the logic: "If there is **anything** not expected - not behaving as it should - not valid => raise an Exception (not just log) and disengage completely variant behavior.

Log error and proceed will lead to a lot of frustration because it will be buried inside builders log.

mgorny:
Updated. That said, I had to remove `@VariantCache()` use since it does not seem support being disabled/flushed, and therefore makes it impossible to test with different plugins.

rgommers:
> "If there is **anything** not expected - not behaving as it should - not valid => raise an Exception (not just log) and disengage completely variant behavior.

Can we define what errors are expected and who raises them? I can imagine that for `variantlib` it isn't always clear what to do - for example, at wheel build time a missing plugin should raise while at install time it may select a non-variant wheel. Hence the tri-state API seems useful.

The type of exception to raise in case `variantlib` does end up doing the raising, and making that exception part public API, will also be useful. A frontend/backend may for example want to catch and re-raise to add more context.

mgorny:
In general, I think it's a good idea, though I'm not sure if it necessarily makes sense for these exceptions. In the end, they only indicate that a particular plugin is broken which really shouldn't be happening on end user systems.

DEKHTIARJonathan:
My perspective - and I might be wrong - I want to reduce to an absolute minimum the number of situations where variantlib gets confused and try to guess what to do or tries to recover from a situation by making a guess.

If there is anything "not legal" => panic and stop.

We may end up relaxing this in the future, however I do think it's critical for variantlib to initially be hyper pedantic and raise exceptions left & right as we teach people how to use variants and build experience with variants. 

What exception doesn't matter too much to me. I added errors. ValidationError because it's useful to use as we parse inputs. Feel free to add any exception you think make sense. 

The most important is not to have "silent errors" like a plugin has a problem, you silently deactivate it and continue. This is a huge issue because most people will not see that problem and don't understand why they ended up with the wrong variants. A solid and consistent error explaining you what the problem is and what to do about it, force the user to fix it and allow them to not get a half functional install. This is especially important when people have build scripts, docker builds and that most installers totally capture all outputs.

variantlib/api.py
```diff
@@ -94,3 +96,42 @@ def get_variant_hashes_by_priority(
             yield vdesc.hexdigest
     else:
         yield from []
+
+
+@dataclass
+class VariantValidationResult:
```
DEKHTIARJonathan, line 102: Please move this under `models` directory. I'd like to keep `api.py` as clean as possible

DEKHTIARJonathan (approved):
@mgorny please move the object under `/models` - just so that we are consistent with the structure. And also it does look that this dataclass should be `frozen` - I don't think it should be possible to modify it right ?

You have green light to merge this PR whenever you think it's ready

PR #24, opened 2025-04-01
[Refactor] Polishing a few edges - Preparing for sorting & filtering

DEKHTIARJonathan:
More minor refactoring. I try to separate the update in view of the filtering & sorting PR coming to make the review of that PR easier and changing the least number of files.

Manipulating all these concepts extensively for filtering & sorting. I agree with you @mgorny. Some "names" were not very explicit and self-explanatory.

I do agree that `key/value` concepts are more confusing than they help. Hence the change.
Though I tried to not get too far because the concept is useful.

### Changes

- Addition of the `VariantFeature` model that you can understand it as `nvidia :: cuda` aka sort of equivalent to `nvidia :: cuda :: *`. It won't be exposed as a public API model I believe - I don't think there's a usecase for it.
But it's useful to filter & sort (*e.g. I want the `nvidia :: cuda` feature to be the most important*)

- Renaming: `VariantMeta` to `VariantProperty` to avoid singular/plural confusion

- Renaming: `VariantMeta.key` into `VariantProperty.feature`

- Renaming: `KeyConfig` into `VariantFeatureConfig`

- Renaming: `VariantFeatureConfig.key` into `VariantFeatureConfig.name`


mgorny:
So it's going to be:

    namespace :: feature-name :: feature-value

now?

variantlib/base.py
```diff
@@ -6,12 +6,12 @@
 
 
 @runtime_checkable
-class KeyConfigType(Protocol):
-    """A protocol for key configs"""
+class VariantFeatureConfigType(Protocol):
+    """A protocol for VariantFeatures configs"""
```
mgorny, line 10:
```suggestion
    """A protocol for VariantFeature configs"""
```

variantlib/models/base.py
```diff
@@ -14,3 +19,31 @@ def __post_init__(self) -> None:
...
+
+
+@dataclass(frozen=True)
+class BaseHashableModel(ABC, BaseModel):
+    @abstractmethod
+    def _data_to_hash(self) -> list[bytes]:
+        """
+        Convert the object to a list of bytes that will be used for hashing.
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
```
mgorny, line 31: For the record, I think it is customary to skip the message entirely, since it is never raised anyway.

variantlib/models/variant.py
```diff
@@ -67,69 +122,80 @@ class VariantMeta(BaseModel):
...
+            return NotImplemented
+
+        return (
+            self.namespace == other.namespace
+            and self.feature == other.feature
+            and self.value == other.value
+        )
+
+    @property
+    def property_hexdigest(self) -> str:
```
mgorny, line 136: Is this hash actually going to be used anywhere public? If not, then there's really no point in adding extra complexity and abstraction (= making code less obvious), if `hash()` would suffice for internal needs.

variantlib/models/base.py
DEKHTIARJonathan: Oh right - I'll remove it ;)

variantlib/models/variant.py
DEKHTIARJonathan: Let me try to replace with `hash()` - it should indeed work :)

DEKHTIARJonathan: Fixed in 3d7d49b514d83f1c8a83734dfbba720b9393d1f5

variantlib/models/base.py
DEKHTIARJonathan: Fixed in 3d7d49b514d83f1c8a83734dfbba720b9393d1f5

variantlib/models/variant.py
```diff
@@ -44,17 +41,68 @@ class VariantMeta(BaseModel):
...
+                    lambda v: validate_matches_re(v, VALIDATION_FEATURE_REGEX),
                 ],
                 value=val,
             )
         }
     )
+
+    @property
+    def feature_hash(self) -> int:
+        return hash((self.namespace, self.feature))
```
mgorny, line 58: Doesn't dataclass automatically generate a `__hash__()` method here, so you could use `hash(self)`? Then, I don't think we should create a custom hash property when `hash(VariantFeature)` is the usual way of doing this.

variantlib/models/variant.py
```diff
@@ -44,17 +41,68 @@ class VariantMeta(BaseModel):
...
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, VariantFeature):
+            return NotImplemented
+        return self.namespace == other.namespace and self.feature == other.feature
```
mgorny, line 63: This also looks redundant to what `dataclass` creates automatically.

variantlib/models/variant.py
DEKHTIARJonathan: I need a double hash value here for the `VariantProperty`:
- Feature Hash (doesn't count the value)
- Property Hash (does count the value)

variantlib/models/variant.py
DEKHTIARJonathan: Let me try to remove this

DEKHTIARJonathan: Fixed in 348bd2aeeaeb687ef876c7ee1d7a02bbcf7b46a2

variantlib/models/variant.py
mgorny: Hmm, I think I'm getting slightly confused about the nomenclature again.

So we have "namespace", "feature name" and "feature" (that's "namespace" + "feature")?

I think that if you need a hash specifically of these two values, it would be clearer to just `hash((property.namespace, property.feature))` or otherwise use `(property.namespace, property.feature)` explicitly as a "unique key".

DEKHTIARJonathan: To avoid using 20+ words we do "aggregate" some similar concepts under the same name.
So when we talk about the `string` representation:
```
property_str = "namespace :: feature :: value"
```

All of that combined represents a `VariantProperty` (which is what we agreed one).

Now `feature` by itself does not convey any meaning. If we wanted to be accurate it would actually be:
```
property_str = "namespace :: feature_name :: feature_value"
```

It's very heavy and verbose but now it's accurate.

So what is unique to a variant is the combination of `namespace + feature_name` (you can't have 2 of those inside one variant). And that is a `VariantFeature.`

We could have create a `VariantNamespace` class only containing the `namespace: str` field. But I don't see the use-case.

So in summary:

```yaml
VariantProperty:
    VariantFeature:
        namespace: str
        feature: str  # aka "feature-name"
    value: str  # aka "feature-value"
```

The whole concept hasn't changed.

mgorny: Yeah, it is clear now. Though I have to say that for someone with C++ background, it would seem more natural if "namespace" and "feature name" combined together into a "name" :-).

variantlib/models/variant.py
```diff
@@ -67,69 +111,63 @@ class VariantMeta(BaseModel):
...
         metadata={
             "validator": lambda val: validate_and(
                 [
                     lambda v: validate_instance_of(v, list),
-                    lambda v: validate_list_of(v, VariantMeta),
+                    lambda v: validate_list_of(v, VariantProperty),
                     lambda v: validate_list_min_len(v, 1),
-                    lambda v: validate_list_all_unique(v, key=hash),
+                    lambda v: validate_list_all_unique(
+                        v, key=attrgetter("feature_hash")
```
mgorny, line 169: I think:

```suggestion
                        v, key=lambda x: (x.namespace, x.key)
```

should work just the same, and avoid unnecessary inheritance. Also, it matches 1:1 what's done below.

DEKHTIARJonathan: Oh interesting. I thought I was using this more :D

DEKHTIARJonathan: Yes in this PR it's easy. Unfortunately I extensively need and use that concept for sorting & filtering. So if you don't mind I'd like to keep it :D

variantlib/models/variant.py
DEKHTIARJonathan: Well if you have a better idea let me know :D

variantlib/models/variant.py
mgorny: Now I'm confused. Sorting by `(x.namespace, x.key)` sorts by strings, whereas sorting by hash makes sort order pretty random.

DEKHTIARJonathan: This is not the sorting part - this is the unicity validation.
This part is the sorting: https://github.com/wheelnext/variantlib/pull/24/files#diff-2739a87bc9b043288bb679e59c1a2200798497f3d166f9e8fdea8d0620cce6d7R188

DEKHTIARJonathan: > Now I'm confused. Sorting by (x.namespace, x.key) sorts by strings, whereas sorting by hash makes sort order pretty random.

For the record. Yes as long as you would use the `hash()` function inside python yes. If you were to use a real hashing function - that would not be the case ;) It would be random but consistent. And only the consistency matters not the order. The problem with `hash()` is that it's seeded at random and not consistent.

But anyway doesn't matter since we are not doing that ;)

PR #25, opened 2025-04-04
[Resolver] Basic Logic for Filtering & Sorting added

DEKHTIARJonathan:


DEKHTIARJonathan:
This PR is linked to the following discussion: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/21

variantlib/resolver/sorting.py
```diff
@@ -0,0 +1,218 @@
...
+    validate_instance_of(property_priorities, list)
+    validate_list_of(property_priorities, VariantProperty)
+
+    _property_priorities = [vprop.property_hash for vprop in property_priorities]
+
+    # if not present push at the end
+    try:
+        return _property_priorities.index(vprop.property_hash)
+    except ValueError:
+        return sys.maxsize
```
mgorny, line 40: Interesting idea. Not sure how clean it is, but I don't see any immediate issues.

tests/resolver/test_filtering.py
```diff
@@ -0,0 +1,405 @@
...
+def vdescs(vprops: list[VariantProperty]) -> list[VariantDescription]:
+    """Fixture to create a list of VariantDescription objects."""
+    assert len(vprops) == 2
+    prop1, prop2 = vprops
+
+    return [
+        VariantDescription([prop1]),
+        VariantDescription([prop2]),
+        VariantDescription([prop1, prop2]),
+    ]
```
mgorny, line 39: Perhaps make it more universal:

```suggestion
    return list(itertools.chain.from_iterable(itertools.combinations(vprops, x + 1) for x in range(len(vprops))))
```

tests/resolver/test_filtering.py
```diff
@@ -0,0 +1,405 @@
...
+
+def test_remove_duplicates(vdescs: list[VariantDescription]):
+    # using `copy.deepcopy` to ensure that all objects are actually unique
+    input_vdescs = [copy.deepcopy(random.choice(vdescs)) for _ in range(100)]
+    filtered_vdescs = list(remove_duplicates(input_vdescs))
+
+    assert len(filtered_vdescs) == 3
+
+    for vdesc in vdescs:
+        assert vdesc in filtered_vdescs
```
mgorny, line 53: Wouldn't

```suggestion
    assert sorted(filtered_vdescs) == sorted(vdescs)
```

work? Comparing lists directly should give better diffs.

tests/resolver/test_filtering.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+# ====================== `filter_variants_by_features` ====================== #
+
+
+def test_filter_variants_by_features(vdescs: list[VariantDescription]):
+    vdesc1, vdesc2, _ = vdescs
+
+    vfeat1 = VariantFeature(namespace="OmniCorp", feature="custom_feat")
+    vfeat2 = VariantFeature(namespace="TyrellCorporation", feature="client_id")
```
mgorny, line 182: I wonder if we should a helper to convert `VariantProperty` into a `VariantFeature`. In #32 I've inlined similar logic as well.

tests/resolver/test_filtering.py
```diff
@@ -0,0 +1,405 @@
...
+                allowed_features="not a list",  # type: ignore[arg-type]
+            ),
+            maxlen=0,
+        )
+
+    with pytest.raises(ValidationError):
+        deque(
+            filter_variants_by_features(
+                vdescs=vdescs,
+                allowed_features=["not a `VariantFeature`"],  # type: ignore[list-item]
```
mgorny, line 265: Perhaps also test for passing `VariantProperty` as `allowed_features`, and below for passing `VariantFeature` as `allowed_properties`.

tests/resolver/test_lib.py
```diff
@@ -0,0 +1,389 @@
...
+        VariantDescription([prop3, prop2]),
+        VariantDescription([prop1, prop4]),  # duplicate with prop4
+        VariantDescription([prop4, prop1]),  # duplicate with prop4
+        VariantDescription([prop2, prop4]),  # duplicate with prop4
+        VariantDescription([prop4, prop2]),  # duplicate with prop4
+        VariantDescription([prop1]),
+        VariantDescription([prop2]),
+        VariantDescription([prop3]),
+        VariantDescription([prop4]),
+    ]
```
mgorny, line 69:
```suggestion
    return list(y for y in itertools.chain.from_iterable(itertools.permutations(vprops, x+1) for x in range(len(vprops))) if not (prop3 in y and prop4 in y))
```

EDIT: I've just realized you're depending on specific order here, while `itertools.permutations()` gives a different (but stable) ordering.

tests/resolver/test_lib.py
```diff
@@ -0,0 +1,389 @@
...
+
+    assert sort_and_filter_supported_variants(
+        vdescs=vdescs,
+        supported_vprops=vprops,
+        property_priorities=[vprop1],
+        feature_priorities=[vfeat1],
+        namespace_priorities=["OmniCorp", "TyrellCorporation"],
+    ) == [
+        # everything that contains `vprop2` with vfeat1 important and OmniCorp first
+        vdescs[0],  # 0
```
mgorny, line 341: To be honest, this is quite unreadable. Perhaps inlining the `VariantDescription`s here would be a better idea?

tests/test_configuration.py
```diff
@@ -0,0 +1,151 @@
...
+from variantlib.configuration import ConfigEnvironments
+from variantlib.configuration import VariantConfiguration
+from variantlib.configuration import get_configuration_files
+from variantlib.constants import CONFIG_FILENAME
+from variantlib.models.configuration import VariantConfiguration as ConfigurationModel
+from variantlib.models.variant import VariantFeature
+from variantlib.models.variant import VariantProperty
+
+
+def dict_to_toml(d: dict, indent: int = 0) -> str:
```
mgorny, line 19: Can't we use `tomli_w` here? I realize the output is relatively simple but I feel like implementing our own TOML serializer is going a bit far, especially that it'd be only a test dependency.

tests/test_configuration.py
```diff
@@ -0,0 +1,151 @@
...
+            lines.append(f"{' ' * indent}{key} = [{items}]")
+        elif isinstance(value, str):
+            lines.append(f'{" " * indent}{key} = "{value}"')
+        else:
+            raise TypeError(f"Unsupported type for key '{key}': {type(value)}")
+    return "\n".join(lines)
+
+
+def test_reset():
+    VariantConfiguration._config = ConfigurationModel.default()  # noqa: SLF001
```
mgorny, line 36: Perhaps we should disable `SLF001` entirely for tests?

tests/test_configuration.py
```diff
@@ -0,0 +1,151 @@
...
+        ConfigEnvironments.VIRTUALENV: Path("/nonexistent/config.toml"),
+        ConfigEnvironments.USER: Path("/nonexistent/config.toml"),
+        ConfigEnvironments.GLOBAL: Path("/nonexistent/config.toml"),
+    }
+    config = VariantConfiguration.get_config()
+    assert config == ConfigurationModel.default()
+
+
+@patch("variantlib.configuration.get_configuration_files")
+def test_get_default_config_with_no_file(mock_get_config_files):
```
mgorny, line 76: This seems the same as `test_get_config_no_files`.

variantlib/configuration.py
```diff
@@ -0,0 +1,123 @@
...
+
+class ConfigEnvironments(IntEnum):
+    LOCAL = 1
+    VIRTUALENV = 2
+    USER = 3
+    GLOBAL = 4
+
+
+@cache
+def get_configuration_files() -> dict[ConfigEnvironments, Path]:
```
mgorny, line 44: I think it'd be good to add a test for the values returned here. We could mock `Path.cwd()` and `sys.prefix` to get stable values, and we should be able to affect platformdirs on non-Windows via setting `XDG_CONFIG_DIRS` etc.

variantlib/configuration.py
```diff
@@ -0,0 +1,123 @@
...
+        for config_name in ConfigEnvironments:
+            if (cfg_f := config_files[config_name]).exists():
+                logger.info("Loading configuration file: %s", config_files[config_name])
+                with cfg_f.open("rb") as f:
+                    try:
+                        config = tomllib.load(f)
+                    except tomllib.TOMLDecodeError as e:
+                        raise errors.ConfigurationError from e
+
+                return ConfigurationModel.from_toml_config(**config)
```
mgorny, line 105: I do wonder if we shouldn't be combining configurations. For example, the user could generally prefer `x86-64-v2` in their user configuration, but the current project may be adding `cuda12.6` priority — in which case the resulting order would be `[cuda12.6, x86-64-v2]` rather than ignoring the user config.

variantlib/constants.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 VARIANT_HASH_LEN = 8
-CONFIG_FILENAME = "wheelvariant.toml"
+CONFIG_FILENAME = "variants.toml"
```
mgorny, line 4: Not sure about this change. In the end, it kinda assumes that we will always be dealing with wheel-oriented packages and nothing else would use the same filename. For example, rattler-build currently uses `variants.yaml` which is pretty close to colliding.

variantlib/resolver/filtering.py
```diff
@@ -0,0 +1,176 @@
...
+
+    seen = set()
+
+    def _should_include(vdesc: VariantDescription) -> bool:
+        """
+        Check if any of the namespaces in the variant description are not allowed.
+        """
+        validate_type(vdesc, VariantDescription)
+
+        if vdesc.hexdigest in seen:
```
mgorny, line 32: We should make `VariantDescription` hashable, and use Python hash for this, rather than the slow `.hexdigest` method. Let's not reinvent the wheel unnecessarily.

variantlib/resolver/lib.py
```diff
@@ -0,0 +1,141 @@
...
+    # - Easier to maintain
+    # - Easier to extend
+    # - Easier to test
+    # - Allows `tooling` to provide CLI/configuration flags to explicitly
+    #   reject a `namespace` or `namespace::feature` without complex
+    #   processing.
+    # ======================================================================= #
+
+    # Step 1: Remove duplicates - should never happen, but just in case
+    result = remove_duplicates(vdescs)
```
mgorny, line 65: This seems inconsistent with the policy of throwing exceptions when something that should never happen happens.

variantlib/resolver/lib.py
```diff
@@ -0,0 +1,141 @@
...
+
+    yield from result
+
+
+def sort_and_filter_supported_variants(
+    vdescs: list[VariantDescription],
+    supported_vprops: list[VariantProperty] | None = None,
+    property_priorities: list[VariantProperty] | None = None,
+    feature_priorities: list[VariantFeature] | None = None,
+    namespace_priorities: list[str] | None = None,
```
mgorny, line 97:
```suggestion
    namespace_priorities: list[str] | None = None,
    feature_priorities: list[VariantFeature] | None = None,
    property_priorities: list[VariantProperty] | None = None,
```

Other functions use `(namespace, feature, value)` ordering, so here a similar ordering (with `namespace_priorities` first) seems more consistent.

variantlib/resolver/sorting.py
```diff
@@ -0,0 +1,210 @@
...
+        return namespace_priorities.index(vprop.namespace)
+    except ValueError:
+        return sys.maxsize
+
+
+def get_variant_property_priority_tuple(
+    vprop: VariantProperty,
+    property_priorities: list[VariantProperty] | None,
+    feature_priorities: list[VariantFeature] | None,
+    namespace_priorities: list[str] | None,
```
mgorny, line 95:
```suggestion
    namespace_priorities: list[str] | None,
    feature_priorities: list[VariantFeature] | None,
    property_priorities: list[VariantProperty] | None,
```

Likewise.

mgorny (commented):
Hm, do I understand correctly that this currently only accounts for explicit namespace, feature, property priorities specified by user, but doesn't define fallback feature and property priorities based on the order returned by plugins?

tests/resolver/test_filtering.py
DEKHTIARJonathan: It's equivalent but a lot less readable and hard to understand in just a quick look.
I will not merge this - I prefer keeping the readability a tad better at the expense of a little more code.

tests/resolver/test_filtering.py
DEKHTIARJonathan: Does not work because `VariantDescription` does not support ordering. And it's a good thing. The whole logic of this PR is to provide a sorting interface conditional of a variant provider sorted list.
Adding a naive ordering could be confusing (and lead to silent bugs).

tests/resolver/test_filtering.py
DEKHTIARJonathan: Addressed in b4295f8af7ab9da9e7391f01a574d40b7e44f66e

variantlib/resolver/lib.py
DEKHTIARJonathan: Policy of throwing exceptions whenever smthg weird happens happens totally.
However, I don't think in this case it makes too much sense... If the same Variant is passed 20 times or 1 time the result will be the same if I try to sort (even if I don't remove them). Let's say that `aaaa1111` is the best variant. Whether `aaaa1111` is `priority[0]` or `priority[0:20]` doesn't change anything given that we focus on the "best variant" => position 0.

There's another of these usecases where we should "bend the rule":
=> we are not going to force the user to define a namespace priority if they only have one plugin installed. We can be "smart enough" to assume that if there's only one namespace enabled we don't need to set who is the best/worst -> there's only one.

I think the rule that you mention is probably "best formulated" as => **Making absolutely zero guess**.
- Either you provide me a complete configuration / settings that allows me to make a fully deterministic decision.
- Or the situation commands itself that I can "infer fully deterministically without a doubt" what you intend on doing.

variantlib/resolver/lib.py
DEKHTIARJonathan: It's fine with me - just would like to mention smthg and confirm you still agree with what you suggest above:

properties > features > namespace>

See `variants.toml`

```toml
# ======================= EXAMPLE `variants.toml` FILE ====================== #

# ============= Top-Most Priority ============= #

# 1. Define the priority of variant properties - 1st order priority.
#    => Expected format: "namespace::feature::value"
#
# OPTIONAL: In most cases - users will not specify the following. Only used when the user wants
#           to modify the default VariantProperty priority ordering.
#
# Note: This is a lazy-list: Only specify the ones you want to "bump up" to the top of the list

property_priority = [
    "fictional_hw::architecture::mother",
    "fictional_tech::risk_exposure:25",
]

# ============= Second-Most Priority ============= #

# 2. Define the priority of variant features - 2nd order priority under the variant properties.
#    => Expected format: "namespace::feature"
#
# OPTIONAL: In most cases - users will not specify the following. Only used when the user wants
#           to modify the default VariantFeature priority ordering.
#
# Note: This is a lazy-list: Only specify the ones you want to "bump up" to the top of the list

features_priority = [
    "fictional_hw::architecture",
    "fictional_tech::risk_exposure",
    "simd_x86_64::feature3",
]

# ============= Default Priority Ordering ============= #

# 3. Define the priority of variant namespaces
#    => Expected format: "namespace"
#
# MANDATORY AND IMPORTANT - PLEASE READ !
#    - As long as there is more than 1 variant plugin installed. This field must be specified.
#      => no default ordering is assumed.
#
#    - If len(plugins) > 1 and an installed plugin is missing in the priority list => raise ConfigurationError
#
#    - If namespaces is specified by no plugin uses it => warning issued.

namespaces_priority = [
    "fictional_hw",
    "fictional_tech",
    "simd_x86_64",
    "non_existent_provider",
]
```

Do you still want to keep the ordering of arguments `(namespace, features, properties)` despite that in this case the "top priority" is property not namespace ?

variantlib/resolver/sorting.py
DEKHTIARJonathan: Same as above

variantlib/resolver/lib.py
mgorny: Yeah, I think consistency in argument order (even if they are not strictly the same arguments) is more important — at least I would expect to pass namespace first. You could also look at it as passed in order from lowest priority — first we sort by namespace priorities, then by feature priorities, then by property priorities.

DEKHTIARJonathan: That works for me. Changing

variantlib/constants.py
DEKHTIARJonathan: FYI I searched the entire github - There is not one one file called `variants.toml` anywhere ...
Hence I simplified ;)

mgorny: Actually, thinking about it more, perhaps it should be named `variantlib.toml`, since it configures `variantlib`.

variantlib/resolver/filtering.py
DEKHTIARJonathan: It really doesn't matter - Going completely **EXTREME**:

```python
import random
import string
import timeit

from variantlib.models.variant import VariantDescription
from variantlib.models.variant import VariantProperty
from variantlib.resolver.filtering import remove_duplicates


def random_str(length):
    """Generates a random string of specified length.

    Args:
      length: The length of the string to generate.

    Returns:
      A random string of the specified length.
    """

    characters = string.ascii_letters + string.digits
    return "".join(random.choice(characters) for i in range(length))


if __name__ == "__main__":
    vdescs = [
        VariantDescription(
            properties=[
                VariantProperty(
                    namespace=random_str(8),
                    feature=random_str(8),
                    value=random_str(8),
                )
                for _ in range(1000)
            ],
        )
        for _ in range(1000)
    ]

    print("Running benchmarks ...")
    total_time = timeit.timeit(
        lambda: list(remove_duplicates(vdescs)),
        number=10_000,
    )

    print(f"Total time for 10k iters: {total_time * 1000:.2f} msecs")
    print(f"Time per iteration: {total_time * 1000 / 10_000:.2f} msecs")
```

```python
>>> Total time for 10k iters: 3281.88 msecs
>>> Time per iteration: 0.33 msecs
```

There is no world where we have to remove duplicates from more than 1000 variants each with 1000 properties ... 
And even then it takes 0.33 msecs ... Clearly this is good enough

tests/test_configuration.py
DEKHTIARJonathan: I guess for testing it doesn't matter indeed - I didn't want to add dependencies :D

DEKHTIARJonathan: Fixed in e36e93161dc007e0b326bb9d5422111d4bf3192f

tests/test_configuration.py
DEKHTIARJonathan: Not sure how to do that - And it's not in variantlib anyway - Feel free to contribute that to the monorepo

tests/test_configuration.py
DEKHTIARJonathan: Indeed removing

variantlib/resolver/filtering.py
mgorny: It's not really about performance but about doing things the canonical way — not reinventing the wheel when Python has a standard solution, so that when people read the code, they will not have to ask themselves "why is it done like this? what am I missing?"

variantlib/configuration.py
DEKHTIARJonathan: I'm honestly not so sure what you mean with `we should be able to affect platformdirs on non-Windows via setting XDG_CONFIG_DIRS`

I copied the logic from `pip` - nothing more / nothing less.

For the tests - I may let you add it in a following PR - not exactly sure what you have in mind

variantlib/configuration.py
DEKHTIARJonathan: it's possible - we don't need to solve it here - we can refine that logic later.
We won't get it perfect from the first shot no matter what decision we pick today.

variantlib/configuration.py
mgorny: I mean a test that would verify whether we are getting expected paths. Like if we mock user's home to be `/home/user`, then we should get `/home/user/.config/...` and so on.

variantlib/constants.py
DEKHTIARJonathan: let's keep `variants.toml` for now - very easy to change given that it's only in one place in the codebase

tests/resolver/test_filtering.py
DEKHTIARJonathan: I actually added it : https://github.com/wheelnext/variantlib/pull/25/files#diff-2739a87bc9b043288bb679e59c1a2200798497f3d166f9e8fdea8d0620cce6d7R119-R122

tests/resolver/test_lib.py
DEKHTIARJonathan: I can do that :)

PR #27, opened 2025-04-07
Add complete error handling to `PluginLoader`, and cover almost all of it with tests

mgorny:
Now any error to load or instantiate a plugin, or invalid return value will trigger an explicit exception. I've also covered all the cases I could think of.

I've assumed that the entry point is allowed to return any callable that instantiates into `PluginType` class rather than the class type itself.

This also comes with a `validate_type` function that replaces separate `validate_instance_of` and `validate_list_of` calls. The next step would be inferring types straight from dataclasses but that's a followup item, and I'm not sure if it'll be doable cross-module.

I've filed https://github.com/python/mypy/issues/18894 for the `type: ignore`, but I'm not sure if it's feasible to fix on mypy end.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,10 @@ maintainers = [
     { name = "Jonathan Dekhtiar", email = "jonathan@dekhtiar.com" },
     { name = "Michael Sarahan", email = "msarahan@nvidia.com" },
 ]
-dependencies = ["typing-extensions; python_version < '3.11'"]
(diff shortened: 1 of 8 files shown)


DEKHTIARJonathan:
Oh that's cool - I had no idea we could do that :D 
Good one (though hard to read I gotta say)

DEKHTIARJonathan (approved):
Feel free to merge whenever ready

PR #29, opened 2025-04-08
Initial sketch of get_build_setup() API

mgorny:
Initial sketch of API to get variables specific to build environment. The plugin method takes a list of variant properties (variantlib filters them, so that only properties specific to given plugin are passed) and returns a dict[str, str] of variables.  If two or more plugins return the same key, the values are concatenated.

This also includes improvement to validate_type() to handle dict types. I still need to streamline that function, and ideally make it able to handle arbitrary type nesting.

I also need to add tests for invalid values.

TODO: should we retur `dict[str, list[str]]` instead perhaps?  That would probably be nicer to build systems like Meson, and it would also make concatenating cleaner.

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -10,10 +10,14 @@ import pytest
 
 from variantlib.base import PluginType
 from variantlib.base import VariantFeatureConfigType
+from variantlib.base import VariantPropertyType
 from variantlib.errors import PluginError
+from variantlib.errors import PluginMissingError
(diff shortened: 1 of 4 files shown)


mgorny:
Let's review this as-is to get the demo working without messy branches. Worst case, we can change or rip it out entirely later.

DEKHTIARJonathan (approved):
I'm not so sure where it's going - you can merge when you think it's good enough and we'll see later

PR #30, opened 2025-04-08
Separated improvements from #29

mgorny:
Focused on getting the internal improvements in, independently of #29:

- validate `PluginType` manually rather than via `isinstance()`, giving future support for optional methods and better error messages today
- support sorting `VariantFeature` and `VariantProperty` directly
- add unit tests to `validate_type()`
- support validating dicts in `validate_type()`
- support validating recursive types in `validate_types()` (e.g. `dict[str, list[str]]`)

### DIFF
--- a/tests/models/test_validators.py
+++ b/tests/models/test_validators.py
@@ -1,6 +1,5 @@
 import re
 from dataclasses import dataclass
-from types import NoneType
 from typing import Any
 from typing import Protocol
 from typing import Union
@@ -34,7 +33,7 @@ class MyClass(HalfClass):
 @pytest.mark.parametrize(
     ("value", "expected"),
     [
-        (None, NoneType),
+        (None, type(None)),
         (False, bool),
         (11, int),
         (4.5, float),
@@ -58,7 +57,7 @@ def test_validate_type_good(value: Any, expected: type):
 @pytest.mark.parametrize(
     ("value", "expected", "have"),
     [
-        (None, bool, NoneType),
+        (None, bool, type(None)),
         (False, str, bool),
         (11, float, int),
         (4.5, int, float),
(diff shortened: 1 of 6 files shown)


DEKHTIARJonathan (approved):

DEKHTIARJonathan:
Just to check do you mind verifying if before/after this PR the hash of a specific Variant Description is the same? I'm 99% sure that it is the case. Not that it matters at all - just to make I understand well that it didn't change the behavior.

If you can verify that the entire variants.json in mockhouse still produce the same hashes that would be awesome 😊
If in the weird case it doesn't - I'd just like to understand why (the ordering seems like it should be identical) and document it.

Regardless you can merge. As long as the ordering is consistent it doesnt need to follow any specific order.

** IMPORTANT ** 
For the PEP this is absolutely a standard behavior. If we believe that variants should be sorted by names pace::feature (value is useless you can not have 2 identical features with identical or different values). We need to be extra extra sure this is what we are doing here.

DEKHTIARJonathan (approved):

mgorny:
> If you can verify that the entire variants.json in mockhouse still produce the same hashes that would be awesome

Yep, confirmed. Removed the whole file, got byte-exact result.

> For the PEP this is absolutely a standard behavior. If we believe that variants should be sorted by names pace::feature (value is useless you can not have 2 identical features with identical or different values). We need to be extra extra sure this is what we are doing here.

Yes, that's how it works. It basically treats the dataclasses as a `(namespace, feature)` and `(namespace, feature, value)` tuples for the purpose of sorting. Sorting by value is meaningful when you're dealing with generic property lists (which is not something per se used in the spec, but I suppose it is technically valid).

PR #32, opened 2025-04-11
Support for evaluating "variants" environment specifiers

mgorny:
Add an API to evaluate requirements with environment specifiers such as:

```toml
[build-system]
requires = [
        "variant-cuda; 'cuda :: runtime' in variants",
        "variant-x86-64; 'x86_64' in variants",
        "variant-cuda-extra; 'cuda' in variants and python_version >= '3.11'",
        "no-cuda; 'cuda' not in variants",
]
```

This is meant to be used in pypa/build tool. Currently, it just passes the whole `requires` key to pip/uv — but these tools don't know which variants are enabled. The idea is that build would know that, and preprocess the list, so that pip/uv would get only the dependencies they actually need.

My initial attempt (first commit) tried to use `packaging`, but there were two problems with that: almost entire environment marker API is private/internal, and we would have to implement boolean parser to evaluate compound statements correctly. So instead I went for a few regular expressions that should be reliable enough within the narrow scope of environment markers, and that replace the relevant markers with "always true" or "always false" expressions that pip/uv can handle.

Related: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/15

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ maintainers = [
 ]
 dependencies = [
     "importlib-metadata; python_version < '3.10'",
+    "packaging",
     "typing-extensions; python_version < '3.11'",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ maintainers = [
 ]
 dependencies = [
     "importlib-metadata; python_version < '3.10'",
-    "packaging",
     "typing-extensions; python_version < '3.11'",
 ]
 
--- a/tests/test_envspec.py
+++ b/tests/test_envspec.py
@@ -1,5 +1,7 @@
 import pytest
 
+from variantlib.envspec import FALSE_STR
+from variantlib.envspec import TRUE_STR
 from variantlib.envspec import evaluate_variant_requirements
 from variantlib.errors import InvalidVariantEnvSpecError
 from variantlib.models.variant import VariantDescription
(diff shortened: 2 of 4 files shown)


DEKHTIARJonathan:
I'm a bit lost what is being done here and why... If you can try to explain it would be helpful to review :) I'm probably lacking context

DEKHTIARJonathan (commented):
I'm a bit lost what is being done here and why... If you can try to explain it would be helpful to review :) I'm probably lacking context

mgorny:
Perhaps the following will provide useful context:
- https://github.com/wheelnext/build/pull/1 direct use
- https://github.com/mgorny/numpy/blob/wheel-variants/pyproject.toml#L6 example use for building NumPy

mgorny:
Sorry for my short answer in the morning, as I was in a hurry.

> I'm a bit lost what is being done here and why... If you can try to explain it would be helpful to review :) I'm probably lacking context

Basically, to build variant packages we need plugins installed in the build environment — at the minimum, so that we can verify whether the variants specified are correct. To get them installed, we specify them in `build-system.requires`.

Rather than specifying them unconditionally and effectively forcing all users to install all variant plugins when building from source (i.e. `pip install --no-binary ...`), we make them conditional to variants being specified. This also serves as a prototype for https://github.com/wheelnext/pep_xxx_wheel_variants/issues/15.

What we're working with here is specific interaction between `build` and installers (`pip` or `uv`). When used to build a wheel, `build` creates an isolated build environment, takes the dependencies from `build-system.requires` and PEP517 `get_requires_for_build_wheel` hooks, and passes them over to the installer to prepare the build environment. The dependencies are currently passed along with all the environment specifiers such as `; python_version < '3.11'` and such, and the installers handle that just fine.

However, variant dependencies don't fit exactly into this scheme, since `build` is aware of the variants being built but the installers are not. Now, we have two possible solutions here:

1. We could make installers aware of the variants, so they could handle the environment specifiers appropriately. However, this requires defining an artificial protocol (something like `pip install --variant-for-selector=x86_64::level::v3 setuptools "variant_x86_64 ; 'x86_64' in variants"` — note that this "variant" argument does not select variants in installed packages but merely affects environment selectors for arguments specified directly). It is also harder, which means more work for the demo.
2. We can make `build` "preprocess" the dependencies to evaluate variant selectors, and this is what the pull request provides. In this case, `build` reads the dependencies and passes them to `variantlib`, and `variantlib` finds all variant environment specifiers and replaces them with "true" or "false" values that installers like `pip` and `uv` can consume without being directly aware of the selected variants.

I don't consider this perfect or even "final" implementation — rather the simplest solution that works for demonstrating the particular proposed feature, without marrying us to it, and limiting the costs while it's not clear if that's how we want the problem solved.

rgommers:
> I don't consider this perfect or even "final" implementation — rather the simplest solution that works for demonstrating the particular proposed feature, without marrying us to it, and limiting the costs while it's not clear if that's how we want the problem solved.

For demoing this seems fine indeed. xref https://github.com/wheelnext/pep_xxx_wheel_variants/pull/9/files#r2004342408 for the design discussion

PR #33, opened 2025-04-15
Validator All Unique - Tune Up to make error explicit

DEKHTIARJonathan:
This PR replaces the following:

```python
lambda v: validate_list_all_unique(
    v, key=attrgetter("feature_hash")
)
```

By the following:

```python
lambda v: validate_list_all_unique(
    v, keys=("namespace", "feature")
)
```

The new validator is preferable (though equivalent) to avoid a cryptic error message:
```bash
# Before
variantlib.errors.ValidationError: Duplicate value found: '7153903334020990830' in list.

# After
variantlib.errors.ValidationError: Duplicate value found: '('my_namespace', 'my_feature')' in list.
```

### DIFF
--- a/variantlib/models/validators.py
+++ b/variantlib/models/validators.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import re
 from collections.abc import Iterable
+from operator import attrgetter
 from types import GenericAlias
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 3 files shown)


variantlib/models/validators.py
```diff
@@ -32,21 +33,25 @@ def validate_list_min_len(values: list, min_length: int) -> None:
...
 
     for value in values:
-        if key is not None:
-            value = key(value)  # noqa: PLW2901
+        _value = value
 
-        if value in seen:
-            raise ValidationError(f"Duplicate value found: '{value}' in list.")
+        if keys is not None:
+            _value = tuple([attrgetter(key)(value) for key in keys])
```
mgorny, line 47:
```suggestion
            _value = tuple([getattr(value, key) for key in keys])
```

`attrgetter()` only makes sense when you don't have the value immediately.

mgorny (changes requested):

mgorny (approved):
LGTM

PR #35, opened 2025-04-15
[Command] Addition of a simple command `make_variant`

DEKHTIARJonathan:
This PR adds a command `variantlib make_variant` to manually add variant information "post build" to a wheel.

You can see it as "repackaging a wheel as a variant after a build".

This will be very useful to allow everybody to quickly transform any wheel into a variant without having to modify absolutely anything else.

```bash
variantlib make_variant \
    -f pep_xxx_wheel_variants-1.0.0-py3-none-any.whl \
    -o . 
    -p fictional_tech::quantum::drive \
    -p fictional_hw::platform::TARS
```

After discussing with @warsawnv we thought that this was a critical feature to support initially (as build backends gradually rollout support) and for testing.

This PR also include a "optional extra" called `user` - Just because this command adds a feature on `wheel` which will not be useful in most cases (like for `pip` and I'm not sure how if `pip` and other would like very much having to bring `wheel`), we can change that part later if we think it's not a big deal to add `wheel` as a core dependency.

This PR uses #33

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+user = [
+    "wheel"
+]
 dev = [
     "check-manifest",
     # Pre Commit Hooks
@@ -61,6 +64,7 @@ variantlib = "variantlib.commands.main:main"
 analyze_wheel = "variantlib.commands.analyze_wheel:analyze_wheel"
 analyze_platform = "variantlib.commands.analyze_platform:analyze_platform"
 generate_index_json = "variantlib.commands.generate_index_json:generate_index_json"
+make_variant = "variantlib.commands.make_variant:make_variant"
 
 [tool.pytest.ini_options]
 testpaths = ["tests/"]
--- a/variantlib/commands/make_variant.py
+++ b/variantlib/commands/make_variant.py
@@ -161,12 +161,7 @@ def make_variant(args: list[str]) -> None:
         raise ValueError(f"{input_filepath.name!r} is not a valid wheel filename.")
 
     # Transform properties into a VariantDescription
-    properties_args: list[str] | None = parsed_args.properties
-    if properties_args is None or len(properties_args) == 0:
-        raise ValueError("At least one Variant Property is required.")
-
-    vprops = [VariantProperty.from_str(vprop) for vprop in properties_args]
-    vdesc = VariantDescription(properties=vprops)
+    vdesc = VariantDescription(properties=parsed_args.properties)
 
     with tempfile.TemporaryDirectory() as _tmpdir:
         tempdir = pathlib.Path(_tmpdir)
--- a/variantlib/commands/make_variant.py
+++ b/variantlib/commands/make_variant.py
@@ -1,0 +1,222 @@
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import pathlib
+import re
+import tempfile
+
+import wheel.cli.pack as whl_pck
+from wheel.cli.unpack import unpack as wheel_unpack
+
+from variantlib.api import VariantDescription
+from variantlib.api import VariantProperty
+from variantlib.constants import VARIANT_HASH_LEN
+from variantlib.errors import ValidationError
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def wheel_variant_pack(
+    directory: str | pathlib.Path,
+    dest_dir: str | pathlib.Path,
+    variant_hash: str,
+    build_number: str | None = None,
+) -> str:
+    """Repack a previously unpacked wheel directory into a new wheel file.
+
+    The .dist-info/WHEEL file must contain one or more tags so that the target
+    wheel file name can be determined.
+
+    This function is heavily taken from:
+    https://github.com/pypa/wheel/blob/main/src/wheel/_commands/pack.py#L14
+
+    Minimal changes tried to be applied to make it work with the Variant Hash.
+
+    :param directory: The unpacked wheel directory
+    :param dest_dir: Destination directory (defaults to the current directory)
+    :param variant_hash: The hash of the variant to be stored
+    """
+
+    # Input Validation
+    variant_hash_pattern = rf"^[a-fA-F0-9]{{{VARIANT_HASH_LEN}}}$"
+    if not re.match(variant_hash_pattern, variant_hash):
+        raise ValidationError(f"Invalid Variant Hash Value `{variant_hash}` ...")
+
+    # Find the .dist-info directory
+    dist_info_dirs = [
+        fn
+        for fn in os.listdir(directory)  # noqa: PTH208
+        if os.path.isdir(os.path.join(directory, fn)) and whl_pck.DIST_INFO_RE.match(fn)  # noqa: PTH112, PTH118
+    ]
+    if len(dist_info_dirs) > 1:
+        raise whl_pck.WheelError(
+            f"Multiple .dist-info directories found in {directory}"
+        )
+    if not dist_info_dirs:
+        raise whl_pck.WheelError(f"No .dist-info directories found in {directory}")
+
+    # Determine the target wheel filename
+    dist_info_dir = dist_info_dirs[0]
+    name_version = whl_pck.DIST_INFO_RE.match(dist_info_dir).group("namever")
+
+    # Read the tags and the existing build number from .dist-info/WHEEL
+    wheel_file_path = os.path.join(directory, dist_info_dir, "WHEEL")  # noqa: PTH118
+    with open(wheel_file_path, "rb") as f:  # noqa: PTH123
+        info = whl_pck.BytesParser(policy=whl_pck.email.policy.compat32).parse(f)
+        tags: list[str] = info.get_all("Tag", [])
+        existing_build_number = info.get("Build")
+
+        if not tags:
+            raise whl_pck.WheelError(
+                f"No tags present in {dist_info_dir}/WHEEL; cannot determine target "
+                f"wheel filename"
+            )
+
+    # Set the wheel file name and add/replace/remove the Build tag in .dist-info/WHEEL
+    build_number = build_number if build_number is not None else existing_build_number
+    if build_number is not None:
+        del info["Build"]
+        if build_number:
+            info["Build"] = build_number
+            name_version += "-" + build_number
+
+        if build_number != existing_build_number:
+            with open(wheel_file_path, "wb") as f:  # noqa: PTH123
+                whl_pck.BytesGenerator(f, maxheaderlen=0).flatten(info)
+
+    # Reassemble the tags for the wheel file
+    tagline = whl_pck.compute_tagline(tags)
+
+    # Repack the wheel
+    wheel_path = os.path.join(dest_dir, f"{name_version}-{tagline}-{variant_hash}.whl")  # noqa: PTH118
+    with whl_pck.WheelFile(wheel_path, "w") as wf:
+        logging.info(
+            "Repacking wheel as `%(wheel_path)s` ...", {"wheel_path": wheel_path}
+        )
+        wf.write_files(directory)
+
+    return wheel_path
+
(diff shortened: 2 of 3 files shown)


variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+    if not wheel_info:
+        raise ValueError(f"`{input_filepath.name}` is not a valid wheel filename.")
+
+    # Transform properties into a VariantDescription
+    properties_args: list[str] | None = parsed_args.properties
+    if properties_args is None or len(properties_args) == 0:
+        raise ValueError("At least one Variant Property is required.")
+
+    vprops = [VariantProperty.from_str(vprop) for vprop in properties_args]
+    vdesc = VariantDescription(properties=vprops)
```
mgorny, line 169:
```suggestion
    vdesc = VariantDescription(properties=parsed_args.properties)
```

All of that is handled by the argument parser with my suggestion.

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+
+    with tempfile.TemporaryDirectory() as _tmpdir:
+        tempdir = pathlib.Path(_tmpdir)
+        wheel_unpack(input_filepath, tempdir)
```
mgorny, line 173: I don't think we really need to unpack it. Unfortunately, it looks like `zipfile` lacks proper API to update files in place or copy compressed members (https://github.com/python/cpython/pull/125718), but at least we could avoid writing files to disk and potentially altering their metadata in the process.

I.e. basically open both archives, iterate over members: if not metadata, copy between archives as-is; if metadata, read, alter and write. It will probably even be simpler than the current logic, which seems to make unnecessary changes to the archive (probably because it was adapted from the command altering tags, which we don't do here).

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+
+        dirname = input_filepath.name.split("-py3")[0]
```
mgorny, line 175: This seems to wrongly assume we'd be always dealing with `-py3` wheels, and never `-cp312` etc.

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+            # Move the file pointer to the beginning
+            file.seek(0)
+
+            # Write back the non-empty content
+            file.writelines(lines)
+
+            # Truncate the file to remove any remaining old content
+            file.truncate()
+
+            file.write(f"Variant-hash: {vdesc.hexdigest}\n")
```
mgorny, line 208: This seems to assume that the metadata will never have longdescription. Also, altering metadata should really be done via the email message class (or `packaging.metadata`, but that will fail on new keys).

mgorny (changes requested):

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+
+    parsed_args = parser.parse_args(args)
+
+    wheel_file_re = re.compile(
+        r"""^(?P<namever>(?P<name>[^\s-]+?)-(?P<ver>[^\s-]*?))
+        ((-(?P<build>\d[^-]*?))?-(?P<pyver>[^\s-]+?)-(?P<abi>[^\s-]+?)-(?P<plat>[^\s-]+?)
+        \.whl|\.dist-info)$""",
+        re.VERBOSE,
+    )
```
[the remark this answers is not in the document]
DEKHTIARJonathan: Alright that works with me :) I'll add that 👍

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,222 @@
...
+
+    input_filepath: pathlib.Path = parsed_args.input_filepath
+    output_directory: pathlib.Path = parsed_args.output_directory
+
+    # Input Validation
+    if not input_filepath.is_file():
+        raise FileExistsError(f"Input Wheel File `{input_filepath}` does not exists.")
```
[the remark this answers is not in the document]
DEKHTIARJonathan: Ooops fixed. 
Not sure what to do with `parser.error()`- Feel free to send a follow up PR

variantlib/commands/make_variant.py
DEKHTIARJonathan: hum I need to find a different idea :D

DEKHTIARJonathan: Found it

variantlib/commands/make_variant.py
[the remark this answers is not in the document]
DEKHTIARJonathan: Fixed

variantlib/commands/make_variant.py
DEKHTIARJonathan: I don't disagree though it's a little complicated for a first iteration - feel free to experiment and optimize ;)

DEKHTIARJonathan:
Good enough for a V1 of the command - merging so that a few people can start playing with it.
@mgorny feel free to adjust & optimize the command further :)

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+    Minimal changes tried to be applied to make it work with the Variant Hash.
+
+    :param directory: The unpacked wheel directory
+    :param dest_dir: Destination directory (defaults to the current directory)
+    :param variant_hash: The hash of the variant to be stored
+    """
+
+    # Input Validation
+    variant_hash_pattern = rf"^[a-fA-F0-9]{{{VARIANT_HASH_LEN}}}$"
+    if not re.match(variant_hash_pattern, variant_hash):
```
warsawnv, line 48: If `wheel_variant_pack()` is an internal API, then you control the `variant_hash` and this could probably be an assertion.

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+    dist_info_dirs = [
+        fn
+        for fn in os.listdir(directory)  # noqa: PTH208
+        if os.path.isdir(os.path.join(directory, fn)) and whl_pck.DIST_INFO_RE.match(fn)  # noqa: PTH112, PTH118
+    ]
+    if len(dist_info_dirs) > 1:
+        raise whl_pck.WheelError(
+            f"Multiple .dist-info directories found in {directory}"
+        )
+    if not dist_info_dirs:
```
warsawnv, line 61: Since you're checking the length of a list, and you're explicitly calling `len()` above, it's probably best to also do so here, e.g. `if len(dist_info_dirs) == 0:`

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+    name_version = whl_pck.DIST_INFO_RE.match(dist_info_dir).group("namever")
+
+    # Read the tags and the existing build number from .dist-info/WHEEL
+    wheel_file_path = os.path.join(directory, dist_info_dir, "WHEEL")  # noqa: PTH118
+    with open(wheel_file_path, "rb") as f:  # noqa: PTH123
+        info = whl_pck.BytesParser(policy=whl_pck.email.policy.compat32).parse(f)
+        tags: list[str] = info.get_all("Tag", [])
+        existing_build_number = info.get("Build")
+
+        if not tags:
```
warsawnv, line 75: Another place where it's generally better to use `if len(tags) == 0:`

I generally prefer to be as explicit about the truthiness test as possible.  E.g. if you know you're testing the length of a sequence, use `len`.  If you are testing whether something is `None` or not, use `if foo is None`.  Only when the truthiness or falseiness can be any number of things (e.g. `None`, the empty string, or a zero length sequence) do I typically use a generic `if foo` or `if not foo`.  YMMV!

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+            raise whl_pck.WheelError(
+                f"No tags present in {dist_info_dir}/WHEEL; cannot determine target "
+                f"wheel filename"
```
warsawnv, line 78: I'm kind of surprised you didn't get a linter warning here, since the line 78 string doesn't need to be an f-string.

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+
+    # Set the wheel file name and add/replace/remove the Build tag in .dist-info/WHEEL
+    build_number = build_number if build_number is not None else existing_build_number
+    if build_number is not None:
+        del info["Build"]
+        if build_number:
+            info["Build"] = build_number
+            name_version += "-" + build_number
```
warsawnv, line 87: This one *could* be an f-string: `name_version += f"-{build_number}"`

pyproject.toml
```diff
@@ -34,6 +34,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+user = [
```
warsawnv, line 37: What do you think about using `cli` (or `commands`) for the optional instead?

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+    The .dist-info/WHEEL file must contain one or more tags so that the target
+    wheel file name can be determined.
+
+    This function is heavily taken from:
+    https://github.com/pypa/wheel/blob/main/src/wheel/_commands/pack.py#L14
+
+    Minimal changes tried to be applied to make it work with the Variant Hash.
+
+    :param directory: The unpacked wheel directory
+    :param dest_dir: Destination directory (defaults to the current directory)
```
warsawnv, line 42: Does it?  Or is it required?

variantlib/commands/make_variant.py
```diff
@@ -0,0 +1,212 @@
...
+            if _dir.is_dir() and _dir.name.endswith(".dist-info"):
+                distinfo_dir = _dir
+                break
+        else:
+            raise FileNotFoundError("Impossible to find the .dist-info directory.")
+
+        if not (metadata_f := distinfo_dir / "METADATA").exists():
+            raise FileNotFoundError(metadata_f)
+
+        with metadata_f.open(mode="r+") as file:
```
warsawnv, line 182: Do you think it would be better to write a new file and move it into place atomically?  E.g. copy lines from `METADATA` into `METADATA.new`, then `os.rename()` the latter to the former.  The question I have is, what if this loop errors out or is interrupted in the middle of it?   You can probably also avoid the `seek()` back to the beginning and the `truncate()`.

You probably _still_ need a try/except that deletes the `.new` file and cleans up after itself, reraising the original exception, but at least then you can still recover back to the original state if something unexpected happens.

variantlib/constants.py
```diff
@@ -1,8 +1,21 @@
...
 CONFIG_FILENAME = "wheelvariant.toml"
 
 VALIDATION_NAMESPACE_REGEX = r"^[A-Za-z0-9_]+$"
 VALIDATION_FEATURE_REGEX = r"^[A-Za-z0-9_]+$"
 VALIDATION_VALUE_REGEX = r"^[A-Za-z0-9_.]+$"
+
+METADATA_VARIANT_HASH_HEADER = "Variant-hash"
+METADATA_VARIANT_PROPERTY_HEADER = "Variant"
+
+WHEEL_NAME_VALIDATION_REGEX = re.compile(
```
warsawnv, line 15: A multi-line regex string with comments would help with readability here.

warsawnv (commented):
Thanks for writing this!  I think this may end up being a very common way to create variant wheels.

Comments given; I hope I'm not being too nitpicky!

variantlib/constants.py
mgorny: In general, I agree here but the regex is copied (and slightly modified) from `pip` (and we already have it repeated in too many places).

DEKHTIARJonathan:
Thanks a lot @warsawnv for your comments. Many of the comments you made are actually directly targeting the original implementation from wheel.cli.pack. I'm gonna check tomorrow what is the original code (which I don't want to modify) and which one my code (and I can improve).

I believe in the future some of that code may go inside wheel itself. Let's see.

In any case thanks for the comments, I'll check what I can improve and I'll open a new PR with these changes.

PR #36, opened 2025-04-15
[Command] `generate_index_json` - removing `provider-requires` field

DEKHTIARJonathan:
This PR removes `provider-requires: [namespaceA, namespaceB, etc.]` because it's useless ...

1. The information is already duplicated as part of the variant-info 

2. Nothing is "required" if you don't have the `namespaceA` plugin installed - then you ignore all variants that declare `namespaceA :: anything`

Unless I'm missing the point I think that part is useless or should definitely be renamed to smthg else

### DIFF
--- a/variantlib/commands/generate_index_json.py
+++ b/variantlib/commands/generate_index_json.py
@@ -8,7 +8,6 @@ import logging
 import pathlib
 import zipfile
 
-from variantlib.loader import PluginLoader
 from variantlib.models.variant import VariantProperty
 
 logger = logging.getLogger(__name__)
@@ -53,7 +52,7 @@ def generate_index_json(args: list[str]) -> None:
                 continue
 
             if (variant_hash := wheel_metadata.get("Variant-hash")) is None:
-                logger.info("%s: no Variant-hash", wheel)
+                logger.info("`%s`: not a Wheel Variant - Skipping ...", wheel)
                 continue
             if (variant_entries := wheel_metadata.get_all("Variant")) is None:
                 logger.warning(
@@ -84,24 +83,5 @@ def generate_index_json(args: list[str]) -> None:
                     f"{wheel}: different property assigned to {variant_hash}"
                 )
 
-    all_plugins = PluginLoader.distribution_names
-    provider_requires = set()
-    for namespace in known_namespaces:
-        if (plugin := all_plugins.get(namespace)) is not None:
-            provider_requires.add(plugin)
-        else:
-            logger.warning("No known plugin matches variant namespace: %s", namespace)
-    provider_requires = {
-        plugin
-        for provider in known_namespaces
-        if (plugin := all_plugins.get(provider)) is not None
-    }
-
     with directory.joinpath("variants.json").open("w") as f:
-        json.dump(
-            {
-                "provider-requires": sorted(provider_requires),
-                "variants": known_variants,
-            },
-            f,
-        )
+        json.dump({"variants": known_variants}, f, indent=4, sort_keys=True)


mgorny:
It's not duplicate. If you don't have the plugin installed, you have no way of knowing which package provides the namespace.

mgorny:
> 1. A given namespace may be "provided" by many plugins. Think `SIMD` or `X86_64` ... There will most certainly be forks and many plugins for that specific namespace => how do you chose which one to declare ?

Forks are a corner case, not the default. If you build variant wheels using a particular plugin, you generally expect the users to install them with the same plugin.

The PEP is supposed to make installing variants easier. Ideally, `pip install torch` should do the right thing. However, with what you're proposing it sounds like users will have to:

1. Figure out that a particular package is providing variant wheels in the first place.
2. Do specific research to find out which packages to install to actually make use of variant wheels.
3. Possibly figure out if they actually need to use a fork of some plugin to make things work.

With that approach, we're barely gaining anything over status quo. We don't have a working `pip install torch`, but "please go visit PyTorch website, and execute the pip calls listed there to make things work".

> The variants.json might be generated by the Index (warehouse) and has absolutely no clue that nvidia namespace might be hosted inside nvidia_plugin package.

That is easily solved by including provider packages in the metadata.

DEKHTIARJonathan:
You make good points - if we want to include that specific point of design inside the design of the PEP we need to have consensus on this point.

Which currently we do not.

And as of right now ... The consensus is manual opt-in installation of plugins, absolutely not installation of all plugins in the background automatically.

There are many many problems with making this a standard behavior:

- Some `tools` may consider to do it though I'm not convinced this is something that shall be standardized. What is supposed to happen if package A and package B use the same namespace `SIMD` coming from two different plugins ?

- Also I don't like this idea to force silently this installation of plugins to anyone doing `pip install torch` - Just pytorch may come with 4-5 plugins I dont want all of them installed automatically

- It's mandatory to provide explicit priority between every installed plugin so that `variantlib` is able to take a decision on **who-goes-first** - this will be a real headache if every install command adds more and more plugins.

mgorny:
Automatic installation is tangential to the problem. Even with fully manual installation, the user needs information. You can't really expect users to go out of their to figure out that 1) they need some plugins in the first place, 2) they need to figure out which plugins to install.

> * What is supposed to happen if package A and package B use the same namespace `SIMD` coming from two different plugins ?

This is already a point where the specification falls short, and assuming outright that packages will require conflicting plugins is pretty much burying it immediately. This is pretty much assuming that in order to install package `X`, we will be asking people to do something like: install plugin `Pa`, install dependency `Da`, uninstall plugin `Pa`, install incompatible plugin `Pb`, install dependency `Db`, install package `X` that depends on `Da` and `Db`.

> * Just pytorch may come with 4-5 plugins I dont want all of them installed automatically

You don't. Some users will prefer that over getting a CPU-only PyTorch install by default, and having to go through extra hoops to get a CUDA version. The main point is that isn't *our* choice to make, and if we design the specification specifically to block people from implementing the behavior they prefer, we are either burying the spec or we are asking people to provide incompatible extensions to the spec that will become de facto standards.

mgorny:
If you want it named differently, I'm fine with that. I've named it `requires` because I followed suit with `build-system.requires`. It can be `provider-packages`, if that works better.

DEKHTIARJonathan:
Let's add this `Variant-provider: mynamespace :: mypackage` to the METADATA as agreed.
And then we can revisit this. At least I want to change the name. `build-requires` make sense because you **must** install A, B, C to build. With Variant Providers it's different it's more you **may** install A, B, C

mgorny:
I think we can close this with #40 merged.

PR #37, opened 2025-04-16
Fix make_variant command to use email.parser to process the metadata

mgorny:
Use `email.parser` API to process the wheel metadata correctly, and append variant metadata to it.  The current code incorrectly appended the `Variant*` keys after the body, so they never became part of the metadata.

### DIFF
--- a/variantlib/commands/make_variant.py
+++ b/variantlib/commands/make_variant.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import argparse
+import email.parser
+import email.policy
 import logging
 import os
 import pathlib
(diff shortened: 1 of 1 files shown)


variantlib/commands/make_variant.py
```diff
@@ -179,28 +185,27 @@ def make_variant(args: list[str]) -> None:
...
+        with metadata_f.open(mode="r+b") as file:
+            # Parse the metadata
+            metadata_parser = email.parser.BytesParser()
+            metadata = metadata_parser.parse(file)
 
-            # Remove trailing empty lines
-            while lines and lines[-1].strip() == "":
-                lines.pop()
+            # Replace the variant headers
+            del metadata["Variant"]
```
DEKHTIARJonathan, line 194: Question:
If we `make_variant` an already-variant wheel.

Do you expect to "append new properties" or "replace them" with new properties ?

mgorny: The existing code replaced them, so I followed suit. I can add an option to append instead, but then we'd have to read properties from input, update their values (if we specify the same feature as found in input) and reconstruct `VariantDescription`. Then we also open the can of worms called "should we have an option to delete some property?"

DEKHTIARJonathan (approved):

PR #39, opened 2025-04-16
Make WHEEL_NAME_VALIDATION_REGEX verbose and improve it

mgorny:
1. Make the regex verbose to make it clear what it does.
2. Remove support for `.dist-info` files — we're using the regex on wheels only, so matching `.dist-info` files would give wrong results.
3. Add a convenience `base_wheel_name` group that I'll use in `make_variant` to easily construct filenames.


DEKHTIARJonathan (approved):
For the record - we might get push back if we start modifying that regex beyond the absolutely necessary (i.e. adding the variant hash).

Approving - might have to change course at some point

PR #40, opened 2025-04-16
Add `Variant-provider` metadata

mgorny:
1. Add `Variant-provider` metadata key with dict of packages providing variant namespaces.
2. Add a common `variantlib.api.update_variant_metadata()` function to set variant-specific metadata wholesale in email-style dicts.
3. Use the new metadata in `generate_variants_json` to avoid requiring plugins being installed on index servers.

Also fixing excludes in pre-commit config, as they ignored `generate_variants_json.py` entirely.


DEKHTIARJonathan (changes requested):

DEKHTIARJonathan (approved):
LGTM

PR #41, opened 2025-04-17
Use zipfile to copy files on-the-fly in make_variant

mgorny:
Rather than unpacking the wheel into a temporary directory and then repacking it, use zipfile to stream the contents of the original wheel to the new wheel, altering the METADATA in the process.  This should be more efficient, especially when dealing with large wheels.  It also removes the dependency on `wheel` package.

Ideally, we would copy other files without recompressing them at all, but unfortunately the zipfile module does not offer such a function right now (see https://github.com/python/cpython/pull/125718).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
 
 [project.optional-dependencies]
 cli = [
-    "wheel"
 ]
 dev = [
     "check-manifest",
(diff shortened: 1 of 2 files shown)


DEKHTIARJonathan (changes requested):
Please update the METADATA hash file in RECORD and then LGTM

DEKHTIARJonathan (approved):

PR #43, opened 2025-04-18
Add more tests for `get_configuration_files()`

mgorny:
Add Unix and macOS-specific tests for stability of paths returned by `get_configuration_files()`, and switch to platformdirs' `Path` methods instead of converting the strings to `Path`s manually.


DEKHTIARJonathan (approved):
LGTM

PR #44, opened 2025-04-18
WIP to update public variant sorting API

mgorny:
For now, preparing and cleaning up the code.

### DIFF
--- a/variantlib/api.py
+++ b/variantlib/api.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import logging
-from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from variantlib.combination import filtered_sorted_variants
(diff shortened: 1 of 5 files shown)


variantlib/models/provider.py
```diff
@@ -82,3 +87,9 @@ def pretty_print(self) -> str:
             )
         result_str += f"\n{'#' * 80}\n"
         return result_str
+
+    def to_list_of_properties(self) -> Generator[VariantProperty]:
```
DEKHTIARJonathan, line 91:
```suggestion
    @property
    def sorted_properties(self) -> list[VariantProperty]:
```

it doesn't need to be a generator. We are not talking about 10k objects.
This can easily be made a (cached?) property

variantlib/models/provider.py
```diff
@@ -82,3 +87,9 @@ def pretty_print(self) -> str:
...
+        """Flatten the config into a list of ordered properties"""
+        for feat_cfg in self.configs:
```
DEKHTIARJonathan, line 93:
```suggestion
        result = []
        for feat_cfg in self.configs:
```

variantlib/models/provider.py
```diff
@@ -82,3 +87,9 @@ def pretty_print(self) -> str:
...
+            for value in feat_cfg.values:
+                yield VariantProperty(self.namespace, feat_cfg.name, value)
```
DEKHTIARJonathan, line 95:
```suggestion
                result.append(VariantProperty(self.namespace, feat_cfg.name, value))
        
        return result
```

variantlib/api.py
```diff
@@ -36,71 +36,38 @@
...
+    variants_json: dict,
+) -> Generator[VariantDescription]:
+    def variant_to_vprops(namespaces: dict) -> Generator[VariantProperty]:
+        for namespace, keys in namespaces.items():
+            for key, value in keys.items():
+                yield VariantProperty(namespace=namespace, feature=key, value=value)
+
+    for vhash, namespaces in variants_json["variants"].items():
+        vdesc = VariantDescription(list(variant_to_vprops(namespaces)))
+        assert vhash == vdesc.hexdigest
```
DEKHTIARJonathan, line 49:
```suggestion
        if vhash != vdesc.hexdigest:
            raise ValidationError(f"Inconsistent hash found: `{vhash=}` != `{vdesc.hexdigest=}`")
```

variantlib/api.py
```diff
@@ -36,71 +36,38 @@
 ]
 
 
+def _unpack_variants_json(
+    variants_json: dict,
+) -> Generator[VariantDescription]:
```
DEKHTIARJonathan, line 41: Same not convinced we need a generator here.
We shouldn't abuse that all over the place. It makes the code much more complicated to follow.
Especially that you're not unpacking 1-2 variants from the file. You're unpacking all of them.
I don't see the value here.

DEKHTIARJonathan (changes requested):

DEKHTIARJonathan (approved):

DEKHTIARJonathan:
I'll modify in a follow up PR directly - easier

Issue #23, opened 2025-04-01
[Technical Debt] Unsorted list of "todos" mentioned here and there:


DEKHTIARJonathan:
- [ ] validation could maybe be done with [typing.get_type_hints()](https://docs.python.org/3/library/typing.html#typing.get_type_hints)
https://github.com/wheelnext/variantlib/pull/22#discussion_r2022402587
*Note Jonathan:* let's see how it works - not convinced it's really worth it

- [ ] using `re.VERBOSE` in `regex`: https://github.com/wheelnext/variantlib/pull/22#discussion_r2022405363

PR #45, opened 2025-04-18
[Validation refactor] Moving the validators outside of `models/`

DEKHTIARJonathan:
Moving the validators outside of `models/` given it's used all over the place.

Also added `validate_variant_json` used to validate the format of `variants.json`


variantlib/validators.py
```diff
@@ -153,3 +158,96 @@ def validate_type(value: Any, expected_type: type) -> None:
     wrong_type = _validate_type(value, expected_type)
     if wrong_type is not None:
         raise ValidationError(f"Expected {expected_type}, got {wrong_type}")
+
+
+def validate_variant_json(data: dict) -> None:
```
mgorny, line 163:
```suggestion
def validate_variants_json(data: dict) -> None:
```

I think it's clearer when it matches the filename.

variantlib/validators.py
```diff
@@ -153,3 +158,96 @@ def validate_type(value: Any, expected_type: type) -> None:
...
+
+    if not isinstance(variant_data, dict):
+        raise ValidationError(
+            f"Expected a dictionary for `data['variants']`, got `{type(variant_data)}`."
+        )
+
+    hash_re = re.compile(rf"^[0-9a-f]{{{VARIANT_HASH_LEN}}}$")
+    namespace_re = re.compile(VALIDATION_NAMESPACE_REGEX)
+    feature_re = re.compile(VALIDATION_FEATURE_REGEX)
+    value_re = re.compile(VALIDATION_VALUE_REGEX)
```
mgorny, line 184: Perhaps we should keep compiled versions of these in `variantlib.constants`.

variantlib/validators.py
```diff
@@ -153,3 +158,96 @@ def validate_type(value: Any, expected_type: type) -> None:
...
+
+    for variant_hash, vdata in variant_data.items():
+        # Check if variant_hash is a string and matches the regex
+        if not isinstance(variant_hash, str):
+            raise ValidationError(
+                f"Invalid `variant_hash` type: `{variant_hash}` in data (expected str)"
+            )
+        if re.match(hash_re, variant_hash) is None:
```
mgorny, line 192:
```suggestion
        if hash_re.match(variant_hash) is None:
```

If you compile it, you should use the compiled object.

variantlib/validators.py
```diff
@@ -153,3 +158,96 @@ def validate_type(value: Any, expected_type: type) -> None:
...
+
+        # Check the Variant Data
+        if not isinstance(vdata, dict):
+            raise ValidationError(
+                f"Invalid variant data for hash `{variant_hash}`: {type(vdata)}"
+            )
+
+        if len(vdata) == 0:
+            raise ValidationError(
+                f"Invalid no variant properties declared for `{variant_hash}`"
```
mgorny, line 203: That's non-grammatical.

```suggestion
                f"No variant properties declared for `{variant_hash}`"
```

variantlib/validators.py
```diff
@@ -153,3 +158,96 @@ def validate_type(value: Any, expected_type: type) -> None:
...
+            )
+
+        for namespace, vprop_data in vdata.items():
+            # Check if namespace is a string and matches the regex
+            if not isinstance(namespace, str):
+                raise ValidationError(
+                    f"Invalid variant namespace `{namespace}` for hash `{variant_hash}`"
```
mgorny, line 210: Don't you want to include the type here?

mgorny (commented):
I commented only on the first instance of each problem.

variantlib/validators.py
DEKHTIARJonathan: Agreed

variantlib/validators.py
DEKHTIARJonathan: Merged

variantlib/validators.py
DEKHTIARJonathan: Merged

variantlib/validators.py
DEKHTIARJonathan: True - Doing that - We should do that to other regex too

tests/test_validators.py
```diff
@@ -103,3 +106,37 @@ def test_validate_type_bad(value: Any, expected: type, have: type):
...
+
+
+def test_validate_variants_json_empty():
+    validate_variants_json({"variants": {}})
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        {},
```
mgorny, line 129: Let's cover some pathological cases too.

```suggestion
        {},
        {"variants": ["abcd1234"]},
```

tests/test_validators.py
```diff
@@ -103,3 +106,37 @@ def test_validate_type_bad(value: Any, expected: type, have: type):
...
+        {"variants": {"abcd12345": {}}},
+        {"variants": {"abcd1234": {}}},
```
mgorny, line 131:
```suggestion
        {"variants": {"abcd1234": {}}},
        {"variants": {"abcd1234": ["namespace"]}},
```

variantlib/models/variant.py
```diff
@@ -127,14 +129,16 @@ def to_str(self) -> str:
     @classmethod
     def from_str(cls, input_str: str) -> Self:
         # removing starting `^` and trailing `$`
-        pttn_nmspc = VALIDATION_NAMESPACE_REGEX[1:-1]
-        pttn_feature = VALIDATION_FEATURE_REGEX[1:-1]
-        pttn_value = VALIDATION_VALUE_REGEX[1:-1]
+        pttn_nmspc = VALIDATION_NAMESPACE_REGEX.pattern[1:-1]
+        pttn_feature = VALIDATION_FEATURE_REGEX.pattern[1:-1]
+        pttn_value = VALIDATION_VALUE_REGEX.pattern[1:-1]
```
mgorny, line 134: Hmm… I think we can drop the `^` and `$` bits out of regex and use `.fullmatch()` where necessary instead. But let's look into this separately, I'll add it to #23.

variantlib/validators.py
```diff
@@ -153,3 +158,92 @@ def validate_type(value: Any, expected_type: type) -> None:
...
+            for feature_name, feature_value in vprop_data.items():
+                # Check if feature_name is a string and matches the regex
+                if not isinstance(feature_name, str):
+                    raise ValidationError(
+                        f"Invalid variant feature name `{feature_name}` for hash "
+                        f"`{variant_hash}`"
+                    )
+                if VALIDATION_FEATURE_REGEX.match(feature_name) is None:
+                    raise ValidationError(
+                        f"Invalid variant feature name`{feature_name}` for hash "
```
mgorny, line 234:
```suggestion
                        f"Invalid variant feature name `{feature_name}` for hash "
```

PR #46, opened 2025-04-18
[Plug N Play] All the pieces reconnected together - Top level API functioning ready for integration in PIP

DEKHTIARJonathan:
- Align the names [there was a bunch of name with different forms]:
    - namespace
      - namespace_priority
      - namespace_priorities **=> Aligned to this**
      - namespaces_priority
    - feature
      - feature_priority
      - feature_priorities **=> Aligned to this**
      - features_priority
    - property
      - property_priority
      - property_priorities **=> Aligned to this**
      - properties_priority
      
- The idea that `get_combinations` can output a sorted order is:
1. Flawed in nearly all cases because it's dynamic depending on user inputs/configurations
2. Possible with extreme complexity
3. Completely unnecessary it's a testing it needs to stay simple.

- Moved the `unpack_variants_json` in a separate file (AFAIK not a public API)

- `get_variant_hashes_by_priority` is now feature-complete AFAIK (supports user-supplied priorities / forbidden objects).

- `VariantDescription` objects can now do: `from_dict()` and `to_dict()` => useful to go IN/OUT of `variants.json`
    - used both in `api.py` and `generate_index_json`


tests/test_api.py
```diff
@@ -187,19 +130,33 @@ def test_get_variant_hashes_by_priority_roundtrip(configs):
...
+    namespace_priorities = list({provider_cfg.namespace for provider_cfg in configs})
+    mocker.patch(
+        "variantlib.configuration.VariantConfiguration.get_config"
+    ).return_value = VConfigurationModel(namespace_priorities=namespace_priorities)
+
+    def get_or_skip_combinations() -> Generator[VariantDescription]:
         for i, x in enumerate(get_combinations(configs)):
-            assume(i < 65536)
+            if i >= 65536:
+                pytest.skip("Too many combinations")
```
mgorny, line 141: Why are you changing this? `assume()` helps hypothesis choose better samples, `skip()` just kills the test entirely.

DEKHTIARJonathan: It's apparently deprecated. I got a warning message. Replace it back with assume you'll see the message

mgorny: No, it's not. I'm guessing you got a warning because you added it to `test_get_variant_hashes_by_priority_roundtrip()` which doesn't use Hypothesis, and it makes sense only when you are actually using Hypothesis.

tests/test_api.py
```diff
@@ -44,92 +45,34 @@ def configs(mocked_plugin_loader: type[PluginLoader]):  # noqa: F811
...
-    """Generate all possible combinations of `VariantProperty` given a list of
-    `ProviderConfig`. This function respects ordering and priority provided."""
+def test_get_variant_hashes_by_priority_roundtrip(mocker, configs):
+    """Test that we can round-trip all combinations via variants.json and get the same
+    result."""
 
-    assert isinstance(provider_cfgs, (list, tuple))
-    assert len(provider_cfgs) > 0
-    assert all(isinstance(config, ProviderConfig) for config in provider_cfgs)
+    def get_or_skip_combinations() -> Generator[VariantDescription]:
```
mgorny, line 52: Why do you need that? The test data is static, so the combination count is static too, so you're either never or always skipping the test.

variantlib/commands/generate_index_json.py
```diff
@@ -91,34 +99,45 @@ def generate_index_json(args: list[str]) -> None:
...
             {
-                "providers": providers,
-                "variants": known_variants,
+                VARIANTS_JSON_PROVIDER_DATA_KEY: providers,
+                VARIANTS_JSON_VARIANT_DATA_KEY: {
+                    vhash: vdesc.to_dict() for vhash, vdesc in known_variants.items()
+                },
             },
             f,
+            indent=4,
```
mgorny, line 141: Well, the initial idea is that `variants.json` was supposed to be as small as possible, and this goes in the opposite direction.

variantlib/models/variant.py
```diff
@@ -111,12 +112,12 @@ class VariantProperty(VariantFeature):
...
     )
 
-    @cached_property
+    @property
     def property_hash(self) -> int:
         # __class__ is being added to guarantee the hash to be specific to this class
         return hash((self.__class__, self.namespace, self.feature, self.value))
 
-    @cached_property
+    @property
```
mgorny, line 120: I don't think this is a good use of property. The returned object is not a property (i.e. something belonging to) of the original `VariantProperty` but rather a completely new, independent object. The usual way of constructing new objects in Python is via method and not property.

Also, it would be more consistent to follow the naming scheme of `to_*` since it's converting another type.

variantlib/utils.py
```diff
@@ -1,8 +1,30 @@
...
 
 class classproperty(property):  # noqa: N801
     def __get__(self, cls: Any, owner: type | None = None) -> Any:
         return classmethod(self.fget).__get__(None, owner)()  # type: ignore[arg-type]
+
+
+T = TypeVar("T")
+
+
+def aggregate_user_and_default_lists(
```
mgorny, line 15: This can either mutate `user_list` in place or return mutable `default_list`. You should either copy the lists, or at least note the mutability in the docstring, since it can be confusing, especially if `default_list` ends up being a global/shared list somewhere.

variantlib/variant_file.py
```diff
@@ -0,0 +1,28 @@
...
+from variantlib.constants import VARIANTS_JSON_VARIANT_DATA_KEY
+from variantlib.errors import ValidationError
+from variantlib.models.variant import VariantDescription
+from variantlib.validators import validate_variant_json
+
+logger = logging.getLogger(__name__)
+
+
+def unpack_variants_json(
+    variants_json: dict[str, dict[str, dict]],
```
mgorny, line 14:
```suggestion
    variants_json: dict,
```

`variants.json` can have other top-level keys whose values may not be `dict`s.

mgorny (changes requested):
Furthermore, I disagree with killing sorted `get_combinations()`. Sure, it does not handle all the possible use cases, **because it was never supposed to**.

It is a second implementation of the ordering algorithm for very specific circumstances (namespaces ordered by priority, no feature/value priorities, no filtering). It is valuable as an independent way of fuzzing the sort algorithm, because it is completely independent of the main sort algorithm. If we start making changes into the algorithm, these changes may not happen to surface in fixed test cases, and that is precisely the purpose of fuzzing.

tests/test_api.py
DEKHTIARJonathan: Ahhh that would explain. My bad I'll revert :)

tests/test_api.py
DEKHTIARJonathan: True. Removing

variantlib/commands/generate_index_json.py
DEKHTIARJonathan: OIC. Though does it really matter to save 1kb and make it more readable ? How about an option `--compact` or `--beautify` (depending which is default).

It's quite hard to read/debug when it's all in one line :/

variantlib/models/variant.py
DEKHTIARJonathan: What about something like `pathlib.Path().parent` ? It does also return a new `Path()`

Like `.parent_feature` - I don't like much `feature_object` taking ideas if you have smthg in mind.

If you really want it as a function we can - this API is pretty much internal-only.

variantlib/utils.py
DEKHTIARJonathan: Oh good point. I don't like mutability. Makes bug tracking very hard to track

DEKHTIARJonathan: Fixed

variantlib/variant_file.py
DEKHTIARJonathan: Ah right

DEKHTIARJonathan:
@mgorny I added back the original implementation of `get_combinations` with a little twist (I added `namespace_priority`) to allow some fun with it a little bit. Added some more unittests since I was there.

variantlib/commands/generate_index_json.py
mgorny: I don't mind changing it. Just pointing out that the feeling I got from the initial discussion is that we wanted to add as little space as possible — but perhaps that was because you didn't like the file back then. I've been using `json_pp` to format it when I needed to read it.

variantlib/models/variant.py
mgorny: Yeah, that's a fair point. The stdlib is already inconsistent, so I guess it's fine either way. (but then, `to_str()` could be a property too, and that would be mightily confusing.)

tests/test_api.py
```diff
@@ -22,9 +22,11 @@
 from variantlib.api import get_variant_hashes_by_priority
 from variantlib.api import set_variant_metadata
 from variantlib.api import validate_variant
+from variantlib.constants import VARIANTS_JSON_VARIANT_DATA_KEY
 from variantlib.loader import PluginLoader
 from variantlib.models import provider as pconfig
 from variantlib.models import variant as vconfig
+from variantlib.models.configuration import VariantConfiguration as VConfigurationModel
```
mgorny, line 29: I don't really understand why the rename.

tests/test_api.py
```diff
@@ -44,92 +46,30 @@ def configs(mocked_plugin_loader: type[PluginLoader]):  # noqa: F811
...
-        VariantDescription([val2c]),
-        VariantDescription([val3a]),
-    ]
+    mocker.patch(
+        "variantlib.configuration.VariantConfiguration.get_config"
+    ).return_value = VConfigurationModel(namespace_priorities=namespace_priorities)
 
+    result = set(get_variant_hashes_by_priority(variants_json=variants_json))
+    combination_vhashs = {vdesc.hexdigest for vdesc in combinations}
+    assert len(result) == len(combination_vhashs)
```
mgorny, line 70: Can't we check the ordering here, given we only do namespace priorities?

tests/test_api.py
```diff
@@ -187,19 +127,32 @@ def test_get_variant_hashes_by_priority_roundtrip(configs):
...
     ).return_value = {provider_cfg.namespace: provider_cfg for provider_cfg in configs}
-    assert list(get_variant_hashes_by_priority(variants_json=variants_json)) == [
-        x.hexdigest for x in combinations
-    ]
+
+    result = set(get_variant_hashes_by_priority(variants_json=variants_json))
+    combination_vhashs = {vdesc.hexdigest for vdesc in combinations}
+    assert len(result) == len(combination_vhashs)
+
+    assert result.symmetric_difference(combination_vhashs) == set()
```
mgorny, line 155: Same here.

tests/test_utils.py
```diff
@@ -0,0 +1,151 @@
+from __future__ import annotations
+
+import pytest
+
+from tests.test_plugins import mocked_plugin_loader  # noqa: F401
```
mgorny, line 5: At this point, I think we should move it to pytest. But that's to be done separately, I'll add to #23.

tests/test_utils.py
```diff
@@ -0,0 +1,151 @@
...
+from tests.utils import get_combinations
+from variantlib.api import VariantDescription
+from variantlib.api import VariantProperty
+from variantlib.loader import PluginLoader
+from variantlib.utils import aggregate_user_and_default_lists
+
+
+@pytest.fixture
+def configs(mocked_plugin_loader: type[PluginLoader]):  # noqa: F811
```
mgorny, line 14:
```suggestion
def configs(mocked_plugin_loader: type[PluginLoader]) -> list[ProviderConfig]:  # noqa: F811
```

tests/test_api.py
mgorny: Yep, readded that, and it works.

PR #48, opened 2025-04-22
Add support for local testing via `tox`

mgorny:
Add support for `tox` to `pyproject.toml`, so that we can conveniently test for multiple Python versions locally too.


DEKHTIARJonathan (approved):
LGTM

PR #49, opened 2025-04-22
Move `mocked_plugin_loader` fixture to `conftest.py`

mgorny:
Move the `mocked_plugin_loader` fixture to `conftest.py`, and shared mocked plugin classes into `mocked_plugins` module.  This way we don't have to import the fixture into other test files, and we don't have to workaround ruff false positives.


DEKHTIARJonathan (approved):
LGTM

PR #50, opened 2025-04-22
Use `re.fullmatch()` instead of embedding `^$`

mgorny:
Use `re.fullmatch()` (and the equivalents) to match the whole string
against the regular expression, rather than explicit `^` and `$`.
This has the same result, but at the same time makes the regular
expression patterns reusable without having to strip these symbols.


DEKHTIARJonathan (approved):
LGTM

PR #54, opened 2025-04-24
[`variantlib configuration set LOCAL`] Fully interactive command

DEKHTIARJonathan:
@mgorny I used your PR and re-spinned it. 

**A few points:**

- We don't need a `set` & `create` command - both are doing the same from a high level perspective (worst case it can just be a flag).

- I don't see the point of `--path` or `--environment`, complicates the command when it can just be `variantlib configuration set LOCAL` without any extra flag [maybe I'm missing something]

- I removed `tomlkit` we don't need it - we already have `tomli_w` in the project as `test` I moved it in CLI too.

---------

Try the full experience of the command, I think you will like it - I spent quite a bit of time to adjust it and make it look really nice & intuitive. Let me know what you think.

Try it - it's hard to understand just by reading the code ;)


variants.toml
```diff
@@ -0,0 +1,45 @@
+########## This file has been auto-generated by VariantLib on 2025-04-24 05:24 ##########
```
mgorny, line 1: Did you intend to commit this file?

variantlib/commands/configuration/set.py
```diff
@@ -0,0 +1,250 @@
...
+        ):
+            instructions = dedent("""\
+                Feature Priorities
+                ==================
+
+                Please select the following variant feature, in preference order,
+                that you wish to prioritize above everything else.\n
+                When you are done select EXIT.\n""")
+
+            known_features = [
```
mgorny, line 200: I presume this is a temporary hack.

pyproject.toml
```diff
@@ -76,6 +78,7 @@ plugins = "variantlib.commands.plugins.main:main"
 [project.entry-points."variantlib.actions.configuration"]
 create = "variantlib.commands.configuration.create:create"
```
mgorny, line 79: You apparently removed the command but did not remove the entry point.

mgorny (commented):
So I have a bunch of concerns:

1. I actually wanted to do that as a next step, as an optional UI (perhaps with `--ui` option), though I thought of using a more friendly library like `panel` (from ncurses) or `urwid`.
2. This requires `curses` that's technically optional in Python. However, that's a minor point.
3. However, the code clearly requires a rich terminal, and won't work on dumb terminals (e.g. where you can't write to arbitrary positions on the screen). This will also make testing commands much harder, as stdin/stdout won't do.
4. My UI idea was slightly different — I wanted the user to be able to move items up and down while still seeing them. While I don't mind the idea of choosing in order per se, we should a least show items that were already selected.
5. I haven't tested that part but I'm concerned that with many features we're going to run out of vertical space, and I don't think the code handles that.
6. This doesn't preserve custom priorities that the user had specified if they aren't part of plugins installed right now. We should at least give the user the option to preserve them.
7. This doesn't give user the view of the current configuration. I imagine the user will sometimes just need to add one plugin or want to reorder something, not always start from scratch.
8. My version did preserve the original file structure, including any custom comments (except for comments inside arrays). This version always overwrites the user config with a template.
9. I'm not convinced "abort after 5 attempts" is a good approach. The user can always choose to exit, and the program suddenly decided to die when you've got 2 out of 3 settings right and you're struggling with the third one would be quite annoying.

I admit my UX wasn't particularly good (though I did follow your suggestions), but I don't think we should be scraping "dumb" UI entirely, because there are valid use cases for it. I think we should be adding the ability to choose a different UI as an option — though I wouldn't start from scratch and rather use a well-tested library that handles issues like the need for scrolling. Also, IMHO a better UX would be to let the user move options up and down, and switching on and off, rather than having to reselect them every time.

DEKHTIARJonathan:
No longer needed

PR #58, opened 2025-04-24
[Cleanup] Removing the `logger` we really don't need it

DEKHTIARJonathan:


variantlib/__init__.py
```diff
@@ -1,7 +1,15 @@
...
-from variantlib import logger  # noqa: F401
+__package_name__ = "variantlib"
 
-__version__ = importlib.metadata.version("variantlib")
+logger = logging.getLogger(__package_name__)
+handler = logging.StreamHandler()
+formatter = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+logger.setLevel(logging.INFO)
```
mgorny, line 13: Are you sure we should be using that in a library? Shouldn't the consumers decide how they want things logged?

DEKHTIARJonathan: When you mean `that` do you mean the format ? Or `logging` library ?

mgorny: The logger setup. FWIU in libraries you just create loggers and log with them, and CLI performs the logger setup — formatters and handlers.

DEKHTIARJonathan: If we replace with:
```
logger = logging.getLogger(__package_name__)
logger.setLevel(logging.INFO)
```

INFO logs gets skipped.

Can you try to play with this configuration and execute `variantlib plugins list`
Make sure that `Discovering Wheel Variant plugins...` gets displayed and tell me what you would like to replace the config with

PR #61, opened 2025-04-25
[generate_index_json] Adjustments & Providers supporting list

DEKHTIARJonathan:
Now produces the following:

```json
{
    "providers": {
        "fictional_hw": [
            "provider_fictional_hw"
        ],
        "fictional_tech": [
            "provider_fictional_tech"
        ]
    },
    "variants": {
        "a99ec74f": {
            "fictional_hw": {
                "architecture": "HAL9000",
                "compute_accuracy": "0",
                "compute_capability": "6",
                "humor": "2"
            }
        },
        "aaac1872": {
            "fictional_hw": {
                "architecture": "HAL9000"
            },
            "fictional_tech": {
                "quantum": "FOAM",
                "risk_exposure": "1000000000",
                "technology": "improb_drive"
            }
        },
        "d109b4f7": {
            "fictional_hw": {
                "architecture": "mother",
                "compute_capability": "4"
            }
        }
    }
}
```


variantlib/commands/generate_index_json.py
```diff
@@ -60,64 +74,24 @@ def generate_index_json(args: list[str]) -> None:
                 logger.warning("%s: no METADATA file found", wheel)
                 continue
 
-            if (
-                variant_hash := wheel_metadata.get(METADATA_VARIANT_HASH_HEADER)
-            ) is None:
-                logger.debug("%s: not a Wheel Variant. Skipping ...", wheel)
```
DEKHTIARJonathan, line 66: Replaced by a regex match above - no need to log smthg irrelevant.

variantlib/commands/generate_index_json.py
```diff
@@ -60,64 +74,24 @@ def generate_index_json(args: list[str]) -> None:
...
 
             vprops = [VariantProperty.from_str(vprop) for vprop in variant_entries]
-            for vprop in vprops:
-                known_namespaces.add(vprop.namespace)
 
             try:
                 vdesc = VariantDescription(vprops)
-                if vdesc.hexdigest != variant_hash:
-                    logger.error(
-                        "%(wheel)s has been rejected due to a variant hash mismatch: "
```
DEKHTIARJonathan, line 110: I removed this part - we don't "care" in this command if the hashes matches / mismatches.
Especially that the value inside METADATA is purely informative and does have any direct use.

Let's just compute the hash on the fly and that's it.

This command is "pre-computing what variant is what hash" - this logic may force us to skip a variant that the installer could have picked up.

variantlib/commands/generate_index_json.py
```diff
@@ -126,13 +100,39 @@ def generate_index_json(args: list[str]) -> None:
...
+                    "%(wheel)s: did not declare any `%(key)s`",
+                    {"wheel": wheel, "key": METADATA_VARIANT_PROVIDER_HEADER},
+                )
+
+            for wheel_provider in wheel_providers:
+                ns, _, provider = wheel_provider.partition(", ")
+                ns = ns.strip()
+                provider = provider.strip()
+                if not ns or not provider:
+                    raise ValueError(
```
mgorny, line 122: The crash here is inconsistent with graceful handling of invalid metadata above.

variantlib/commands/generate_index_json.py
```diff
@@ -126,13 +100,39 @@ def generate_index_json(args: list[str]) -> None:
...
+                        f"{wheel}: Invalid {METADATA_VARIANT_PROVIDER_HEADER}: "
+                        f"{wheel_provider}"
+                    )
+                providers[ns].add(provider)
+
+    sorted_providers = {}
+    for key, values in providers.items():
+        sorted_providers[key] = sorted(values)
```
mgorny, line 130:
```suggestion
    sorted_providers = {
        key: sorted(values) for key, values in providers.items()
    }
```

mgorny, line 130: Or alternatively:

```suggestion
    for values in providers.values():
        values.sort()
```

variantlib/commands/generate_index_json.py
```diff
@@ -126,13 +100,39 @@ def generate_index_json(args: list[str]) -> None:
...
 
-    with directory.joinpath("variants.json").open("w") as f:
+    with (directory / "variants.json").open(mode="w") as f:
```
mgorny, line 132: I personally prefer `joinpath()` over adding extra parentheses but that's just a matter of taste.

variantlib/commands/generate_index_json.py
```diff
@@ -126,13 +100,39 @@ def generate_index_json(args: list[str]) -> None:
                 )
                 continue
 
-            if vdesc.hexdigest not in known_variants:
-                known_variants[variant_hash] = vdesc
+            if (vhash := vdesc.hexdigest) not in known_variants:
```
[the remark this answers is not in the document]
DEKHTIARJonathan: In theory - But that won't work because `variantlib`'s resolver will not allow that.

variantlib/commands/generate_index_json.py
```diff
@@ -45,11 +46,24 @@ def generate_index_json(args: list[str]) -> None:
...
     vprop_parser = email.parser.BytesParser(policy=email.policy.compat32)
     known_variants: dict[str, VariantDescription] = {}
-    known_namespaces = set()
-    providers: dict[str, str] = {}
+    providers: dict[str, set[str]] = defaultdict(set)
 
     for wheel in directory.glob("*.whl"):
+        # Skip non wheel variants
+        if (wheel_info := WHEEL_NAME_VALIDATION_REGEX.fullmatch(wheel.name)) is None:
+            raise TypeError(
```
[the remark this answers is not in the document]
DEKHTIARJonathan: Good point. Just logging an error and `continue`

variantlib/commands/generate_index_json.py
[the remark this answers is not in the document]
DEKHTIARJonathan: I prefer using the `only` absolutely true value => the one provided by variantlib.
If there is a mismatch somewhere, it's the responsibility of the person building the wheel.

This command just "describe what it was able to find on the index"

variantlib/commands/generate_index_json.py
DEKHTIARJonathan: Managed gracefully. You're right a complete error is not a good idea if we can avoid

variantlib/constants.py
```diff
@@ -8,16 +8,21 @@
...
 VALIDATION_VALUE_REGEX = re.compile(r"[A-Za-z0-9_.]+")
 
-METADATA_VARIANT_HASH_HEADER = "Variant-hash"
-METADATA_VARIANT_PROPERTY_HEADER = "Variant"
-METADATA_VARIANT_PROVIDER_HEADER = "Variant-provider"
-
-WHEEL_NAME_VALIDATION_REGEX = re.compile(
+# VALIDATION_PYTHON_PACKAGE_NAME_REGEX = re.compile(r"[^\s-]+?")
+# Per PEP 508: https://peps.python.org/pep-0508/#names
+VALIDATION_PYTHON_PACKAGE_NAME_REGEX = re.compile(
```
mgorny, line 22: Actually, this wasn't supposed to be a "pure" name. We should permit people specifying version ranges too, even if we don't emit them.

PR #63, opened 2025-04-25
[Feature] Null Variant Support

DEKHTIARJonathan:
This PR introduces the spec [tentatively] of `null-variant`. A variant with no variant-properties.
It has systematically the hash `000000000`.

This `null-variant` should be seen as the `default variant` no matter what hardware/software/drivers/etc. is installed.
However being a variant - it's only visible to people with `variant-enabled` installer.

This can be used to:
- Track what's the percentage of users still using the `non-variant` build
- Give a different "fallback" to users depending if their `installer` is `variant enabled` or not.
- The null variant is always added at the end by the resolver as a last resort "fallback"

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -54,9 +54,11 @@ def test_get_variant_hashes_by_priority_roundtrip(mocker, configs):
 
     namespace_priorities = ["test_namespace", "second_namespace"]
 
-    combinations: list[VariantDescription] = list(
-        get_combinations(configs, namespace_priorities)
-    )
+    # The null-variant is always the last one and implicitly added
+    combinations: list[VariantDescription] = [
+        *list(get_combinations(configs, namespace_priorities)),
+        VariantDescription(),
+    ]
     variants_json = {
         VARIANTS_JSON_VARIANT_DATA_KEY: {
             vdesc.hexdigest: vdesc.to_dict() for vdesc in combinations
@@ -127,7 +129,12 @@ def test_get_variant_hashes_by_priority_roundtrip_fuzz(mocker, configs):
             assume(i < 65536)
             yield x
 
-    combinations: list[VariantDescription] = list(get_or_skip_combinations())
+    # The null-variant is always the last one and implicitly added
+    combinations: list[VariantDescription] = [
+        *list(get_or_skip_combinations()),
+        VariantDescription(),
+    ]
+
     variants_json = {
         VARIANTS_JSON_VARIANT_DATA_KEY: {
             vdesc.hexdigest: vdesc.to_dict() for vdesc in combinations
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -57,7 +57,6 @@ def test_get_variant_hashes_by_priority_roundtrip(mocker, configs):
     # The null-variant is always the last one and implicitly added
     combinations: list[VariantDescription] = [
         *list(get_combinations(configs, namespace_priorities)),
-        VariantDescription(),
     ]
     variants_json = {
         VARIANTS_JSON_VARIANT_DATA_KEY: {
@@ -130,10 +129,7 @@ def test_get_variant_hashes_by_priority_roundtrip_fuzz(mocker, configs):
             yield x
 
     # The null-variant is always the last one and implicitly added
-    combinations: list[VariantDescription] = [
-        *list(get_or_skip_combinations()),
-        VariantDescription(),
-    ]
+    combinations: list[VariantDescription] = [*list(get_or_skip_combinations())]
 
     variants_json = {
         VARIANTS_JSON_VARIANT_DATA_KEY: {
--- a/variantlib/commands/analyze_wheel.py
+++ b/variantlib/commands/analyze_wheel.py
@@ -6,12 +6,26 @@ import pathlib
 import re
 import zipfile
 
+from variantlib.constants import METADATA_VARIANT_HASH_HEADER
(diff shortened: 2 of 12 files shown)


tests/test_api.py
```diff
@@ -54,9 +54,11 @@ def test_get_variant_hashes_by_priority_roundtrip(mocker, configs):
 
     namespace_priorities = ["test_namespace", "second_namespace"]
 
-    combinations: list[VariantDescription] = list(
-        get_combinations(configs, namespace_priorities)
-    )
+    # The null-variant is always the last one and implicitly added
+    combinations: list[VariantDescription] = [
+        *list(get_combinations(configs, namespace_priorities)),
```
mgorny, line 59: Why not modify `get_combinations()` to include it?

variantlib/commands/analyze_wheel.py
```diff
@@ -58,22 +72,33 @@ def analyze_wheel(args: list[str]) -> None:
...
+        # Extract all variant provider strings``
+        providers = [
+            ProviderPackage.from_str(provider_str)
+            for provider_str in re.findall(
+                rf"{METADATA_VARIANT_PROVIDER_HEADER}: (.+)", metadata_str
+            )
+        ]
+
+        for line in pretty_print(vdesc=vdesc, providers=providers).splitlines():
+            logger.info(line)
```
mgorny, line 104: By the way, we should really switch to `sys.stdout` here and ideally output things in (semi-)machine-readable way.

variantlib/commands/generate_index_json.py
```diff
@@ -59,13 +60,17 @@ def generate_index_json(args: list[str]) -> None:
...
             logger.debug(
-                "Filepath: `%(input_file)s` ... is not a wheel variant.",
+                "Filepath: `%(input_file)s` ... is not a wheel variant. Skipping ...",
                 {"input_file": wheel.name},
             )
             continue
 
+        if vhash == "0" * VARIANT_HASH_LEN:
+            known_variants[vhash] = VariantDescription()
+            continue
```
mgorny, line 72: Why do we need to special-case it here? I think it would make sense for the algorithm to be able to handle it naturally.

variantlib/models/variant.py
```diff
@@ -194,11 +193,23 @@ def __post_init__(self) -> None:
         # Execute the validator
         super().__post_init__()
 
+    @property
+    def is_null_variant(self) -> bool:
+        """
+        Check if the variant is a null variant.
+        A null variant is a variant with no properties.
+        """
+        return self.hexdigest == "0" * VARIANT_HASH_LEN
```
mgorny, line 202: This seems to be checking a secondary property. I think that by definition, a "null variant" has no properties, so that's what we should check here.

```suggestion
        return not self.properties
```

variantlib/models/variant.py
```diff
@@ -194,11 +193,23 @@ def __post_init__(self) -> None:
...
+
     @cached_property
     def hexdigest(self) -> str:
         """
         Compute the hash of the object.
         """
+        if not self.properties:
```
mgorny, line 209:
```suggestion
        if self.is_null_variant:
```

variantlib/models/variant.py
```diff
@@ -194,11 +193,23 @@ def __post_init__(self) -> None:
         # Execute the validator
         super().__post_init__()
 
+    @property
+    def is_null_variant(self) -> bool:
```
mgorny, line 197: By the way, I really dislike the idea of starting property name with questions like "is". Questions are active part of language, so they naturally suggest a method.

variantlib/validators.py
```diff
@@ -184,9 +185,18 @@ def validate_variants_json(data: dict) -> None:
             raise ValidationError(
                 f"Invalid `variant_hash` type: `{variant_hash}` in data (expected str)"
             )
+
         if VALIDATION_VARIANT_HASH_REGEX.fullmatch(variant_hash) is None:
             raise ValidationError(f"Invalid variant hash `{variant_hash}` in data")
 
+        if variant_hash == "0" * VARIANT_HASH_LEN:
```
mgorny, line 192: Don't we need to check the converse now too?

tests/test_api.py
DEKHTIARJonathan: Ah sure 😅 I didn't think about it 👌

variantlib/commands/analyze_wheel.py
DEKHTIARJonathan: I can do that here that's an easy fix

variantlib/commands/generate_index_json.py
DEKHTIARJonathan: Indeed. Fixed

variantlib/models/variant.py
DEKHTIARJonathan: OK for me - change to methods

variantlib/validators.py
DEKHTIARJonathan: I'm not sure I understand what you said.

mgorny: We're checking that null variant has no properties. Shouldn't we check that regular variant has at least one property?

DEKHTIARJonathan: Ah that makes sense.
I used to rely on the fact that VDesc validator would take care of that.
But for obvious reasons... Not anymore 😂

DEKHTIARJonathan: Done

PR #67, opened 2025-04-30
Use more correct initial ordering in `variantlib config setup`

mgorny:
Correct the initial ordering of options in `variantlib config setup` so that it matches better the order used by `variantlib`, rather than using lexical order.  More specifically:

1. Enabled namespaces are listed first, in their previous configuration order.  The remaining namespaces are sorted lexically.

2. Enabled features are listed first, in their previous configuration order.  The remaining features are sorted according to namespace priorities (respecting the new configuration) and then their plugin ordering (rather than lexically, as previously).

3. Enabled properties are listed first, in their previous configuration order.  The remaining properties are sorted using the resolved sort function, respecting the new namespace and feature priorities, and the plugin ordering of values.


DEKHTIARJonathan (approved):
LGTM

PR #69, opened 2025-05-01
Remove `variantlib.envspec`

mgorny:
Remove `variantlib.envspec` module, as per the new design variant providers will be declared in a separate `pyproject.toml` key and therefore we won't need special processing for `build-requires`.

Related to https://github.com/wheelnext/pep_xxx_wheel_variants/issues/35


DEKHTIARJonathan (approved):
LGTM

PR #73, opened 2025-05-02
Streamline property and requires/entry point regexps

mgorny:
Move the feature and property string validation regular expressions into `variantlib.constants` (I'm going to need them for `pyproject.toml`).


DEKHTIARJonathan (approved):
LGTM

PR #75, opened 2025-05-02
Add support for parsing and validating `pyproject.toml`

mgorny:
Comes with a fancy validator class. I didn't add tests for the class itself since `test_pyproject_toml.py` tests it thoroughly.


DEKHTIARJonathan (approved):
LGTM

PR #76, opened 2025-05-02
Fix configuration tests not to be affected by system config

mgorny:
Fixes `test_class_properties_with_default` failing when variantlib is configured on the system.


DEKHTIARJonathan (approved):
LGTM

PR #77, opened 2025-05-03
[Vendoring] PyPA/Build Env Logic

DEKHTIARJonathan:
I added the core `build.env` logic as vendored files into this PR.

We will need this to later auto-install plugins.

I saw that you contributed to `pypa/build` so I would think you are pretty familiar with that part of the code which should help to speed up our development.

I also don't think it's really necessary to re-implement the wheel. This implementation works and is well tested.

I made sure to properly credit where these files come from - leave this original license.
I had to adapt some things to satisfy ruff / mypy - I tried to keep changes to the absolute minimum.

In the future I could imagine some bits end up slightly changing like. It's obviously very "build" oriented.

We can either merge this today or wait to need it - In any case it's ready.
Unless you have a better idea ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
       - name: Install the package and test dependencies
         run: |
           uv venv
+          uv pip install --upgrade pip
           uv pip install -e '.[test]'
       - name: Run tests
         run: uv run --no-project pytest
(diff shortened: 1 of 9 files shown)


mgorny:
Personally, I would wait with this. I'm not 100% sure variantlib is the right place to do it. All frontends (`pip`, `uv`, `build`) naturally have an equivalent logic — perhaps we should look into plugging it so that their own logic is used instead of reinventing the wheel.

DEKHTIARJonathan:
Yes this PR doesn't need to be merged now - It can also be used as a "quick and dirty" way to get to the objective - and later replaced with smthg better.

Let's wait for the need to arise - If we need - it's here

PR #81, opened 2025-05-05
[API Change] Entry-Point to Plugin API

DEKHTIARJonathan:
As agreed by: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/35


variantlib/loader.py
```diff
@@ -35,31 +35,31 @@ class PluginLoader:
...
         try:
-            module = import_module(entry_point_match.group("module"))
-            attr_chain = entry_point_match.group("attr").split(".")
+            module = import_module(plugin_api_match.group("module"))
+            attr_chain = plugin_api_match.group("attr").split(".")
             plugin_callable = reduce(getattr, attr_chain, module)
         except Exception as exc:
             raise PluginError(
-                f"Loading the plugin from entry point {entry_point!r} failed: {exc}"
+                f"Loading the plugin from plugin api {plugin_api!r} failed: {exc}"
```
mgorny, line 50:
```suggestion
                f"Loading the plugin from {plugin_api!r} failed: {exc}"
```

mgorny (commented):
LGTM but let's not do "plugin api" in user-facing messages because it doesn't sound right. I'll change that and merge.

PR #82, opened 2025-05-05
Case Insensitive

DEKHTIARJonathan:
Addresses: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/25

### DIFF
--- a/variantlib/models/variant.py
+++ b/variantlib/models/variant.py
@@ -53,6 +53,18 @@ class VariantFeature(BaseModel):
         }
     )
 
+    def __post_init__(self) -> None:
+        # Normalization of feature values to lowercase
+
+        # Only "legal way" to modify a frozen dataclass attribute post init.
+        if isinstance(self.namespace, str):
+            object.__setattr__(self, "namespace", self.namespace.lower())
+        if isinstance(self.feature, str):
+            object.__setattr__(self, "feature", self.feature.lower())
+
+        # Execute the validator
+        super().__post_init__()
+
     @property
     def feature_hash(self) -> int:
         # __class__ is being added to guarantee the hash to be specific to this class
@@ -104,6 +116,16 @@ class VariantProperty(VariantFeature):
             )
         }
     )
+
+    def __post_init__(self) -> None:
+        # Normalization of feature values to lowercase
+
+        # Only "legal way" to modify a frozen dataclass attribute post init.
+        if isinstance(self.value, str):
+            object.__setattr__(self, "value", self.value.lower())
+
+        # Execute the validator
+        super().__post_init__()
 
     @property
     def property_hash(self) -> int:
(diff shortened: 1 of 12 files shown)


variantlib/models/variant.py
```diff
@@ -202,8 +223,10 @@ def hexdigest(self) -> str:
         #     a :: b :: c
         #     xd :: e :: f
         # would serialize to the same hash.
-        for vprop in self.properties:
-            hash_object.update(f"{vprop.to_str()}\n".encode())
+
+        hash_object = hashlib.sha256(
+            "\n".join([vprop.to_str() for vprop in self.properties]).encode()
+        )
```
mgorny, line 229: Let's not change this. "Newline-separated string" is a weird concept, and having (n-1) properties with a newline, and nth property without is inconsisent.

mgorny (changes requested):
To be honest, I have mixed feelings about this approach. While I agree with the idea that we should accept any input and normalize it, I don't exactly like the idea that data classes implicitly manipulate the data passed to them, and don't preserve the original input.

mgorny:
But let's not block over implementation details.

variantlib/resolver/filtering.py
```diff
@@ -64,6 +64,9 @@ def filter_variants_by_namespaces(
     validate_type(vdescs, Iterable)
     validate_type(forbidden_namespaces, list[str])
 
+    # Normalize to lower case
+    forbidden_namespaces = [ns.lower() for ns in forbidden_namespaces]
```
mgorny, line 68: For the record, I think this is going too far. I can understand normalizing *input* (`pyproject.toml`, metadata, even `variants.json` — though I don't think we have a valid case for the last one), but here we're literally making invalid API calls work.

mgorny:
No, thinking about it further, these changes definitely go too far. As I've said, we should only be normalizing *input* — and probably emitting warnings about non-normalized input while at it. Here, we all putting normalization all over the place: to input, to interim values, to API calls. We're basically hammering the values all over, just in case uppercase appeared somewhere — which implies we aren't doing things correctly, but rather haphazardly straightening out what went wrong.

So if anything, what we should be doing here is:

1. Identifying where non-normalized input can be coming from (actual user input, configuration files, perhaps `pyproject.toml`, wheel metadata, etc.) and normalizing the input there.
2. Rejecting non-normalized data within internal data structures and API calls.

That said, upon further thought I'm starting to think we should really limit this to the bare minimum (like actual user input and configuration files) or not do at all. My points would be:

- `variants.json` should always be generated, and therefore should always have normalized data.
- Wheel metadata is also generated, so it should have normalized values. Technically, we could accept uppercase to maintain compatibility with wheels created prior to the change, but they will be rejected anyway because of the changed hash value.
- `pyproject.toml` is kinda user input but I dare say we want to keep it clean. And since it's entirely new to the spec, there's no concern that we're going to break existing deployments by forcing strict rules.

This leaves us with explicit user input (to variantlib commands) and configuration files. I think it's fine to accept non-normalized values there — but I also don't see much of a point, given that normalization makes things lowercase, so someone actually has to put an actual effort to do things "wrong".

Finally, I should point out that the way that PyPI projects, wheel naming and sdist naming did not enforce normalization from day one has resulted in quite a mess, with different tools applying different rules over time, and the resulting data still being unpredictable today.

DEKHTIARJonathan:
closing in favor of #83

PR #90, opened 2025-05-13
Restore entry points based loader, and use it for CLI commands

mgorny:
- replace `CLIPluginLoader` with `EntryPointPluginLoader` that loads all plugins via entry points
- update the commands to use the new loader
- make `make-variant` automatically determine `plugin-api` from installed plugins
- switch tests back to `ManualPluginLoader` to simplify them


DEKHTIARJonathan (approved):
LGTM

PR #94, opened 2025-05-29
[WIP] Type checking

DEKHTIARJonathan:


variantlib/protocols.py
```diff
@@ -1,12 +1,21 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
```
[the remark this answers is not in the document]
DEKHTIARJonathan: I'll convert to a comment :) 

I just wanted to make it more visible

tests/models/test_provider.py
```diff
@@ -236,16 +238,18 @@ def test_to_list_of_properties():
 
 
 @pytest.fixture
-def valid_namespace():
+def valid_namespace() -> Literal["namespace"]:
```
mgorny, line 241: I would just use `str` here and below. I think the point of using a fixture is to be able to change it without updating all uses.

tests/plugins/test_loader.py
```diff
@@ -51,8 +49,7 @@ def get_supported_configs(self) -> list[VariantFeatureConfigType]:
 class ExceptionPluginBase(PluginType, ABC):
     namespace = "exception_test"
 
-    @abstractproperty
```
mgorny, line 54: Any reason to un-abstract it? Not a big deal but wondering if the type checker had an issue. Also, since there is nothing abstract left here, we should remove `ABC` inherit.

tests/test_api.py
```diff
@@ -82,22 +87,33 @@ def test_get_variant_hashes_by_priority_roundtrip(
     combinations: list[VariantDescription] = [
         *list(get_combinations(configs, namespace_priorities)),
     ]
-    variants_json: dict | VariantsJson = {
+
+    variants_json = {
+        "$schema": VARIANTS_JSON_SCHEMA_URL,
```
mgorny, line 92: Not `VARIANTS_JSON_SCHEMA_KEY`?

variantlib/commands/unmake_variant.py
```diff
@@ -114,7 +114,7 @@ def _unmake_variant(
                         )
                         for line in input_file:
                             rec_filename, sha256, size = line.split(b",")
-                            if rec_filename != variant_dist_info_path:
+                            if rec_filename.decode("utf-8") != variant_dist_info_path:
```
mgorny, line 117: Good catch!

variantlib/plugins/_subprocess.py
```diff
@@ -8,10 +8,16 @@
 from itertools import groupby
 from typing import TYPE_CHECKING
 
-from _variantlib_protocols import PluginType
-from _variantlib_protocols import VariantFeatureConfigType
-from _variantlib_validators_base import ValidationError
-from _variantlib_validators_base import validate_type
+from _variantlib_protocols import PluginType  # type: ignore[import-not-found]
```
mgorny, line 11: Not really a request but I wonder if we could get around this by using regular imports and substituting them while writing.

variantlib/protocols.py
```diff
@@ -52,11 +58,7 @@ def value(self) -> str:
 class PluginType(Protocol):
     """A protocol for plugin classes"""
 
-    @property
-    @abstractmethod
```
mgorny, line 56: We're losing the abstract method integrity check here.

PR #95, opened 2025-06-02
Add a simple API to get `variant.json` for .dist-info metadata

mgorny:
The API is deliberately returning a dict in order to abstract away both filename and contents — technically we could also expand it to create more than a single file in the future.

Fixes wheelnext/pep_xxx_wheel_variants#44


DEKHTIARJonathan:
Does the class "VariantMetadata" still make sense given that we don't have metadata anymore ?

I like the API a lot - simple and efficient

mgorny:
One potential improvement (not sure if it's worth it) is to have a more streamlined way to assert that only a single variant is declared — either a separate `VariantJson` class, or a special property in `VariantsJson`.

DEKHTIARJonathan:
I like it quite a bit ! Reduces complexity.
I consider https://github.com/wheelnext/pep_xxx_wheel_variants/issues/44 closed with this PR.

PR #97, opened 2025-06-09
Support for PEP517-style module plugins and optional `plugin-api`

mgorny:
Per https://github.com/wheelnext/pep_xxx_wheel_variants/issues/38#issuecomment-2894441630 and below.


mgorny:
> Did you remove support for class-base plugin ?
> 
> If not - can you re-add tests for class-base plugins ?

No — and I didn't remove tests for them. The only artifact modified is the one to test installing plugins. The rest of the test suite still uses class-based plugins.

DEKHTIARJonathan (approved):
LGTM

PR #104, opened 2025-06-16
Optional providers, take two

mgorny:
This time adding `optional` boolean to `providers.*`.


tests/artifacts/test-package/pyproject.toml
```diff
@@ -9,7 +9,12 @@ description = "Installable package for testing"
 requires-python = ">=3.9"
 
 [variant.default-priorities]
-namespace = ["installable_plugin"]
+namespace = ["installable_plugin", "non_existing_plugin"]
 
 [variant.providers.installable_plugin]
 requires = ["test-plugin-package"]
+optional = false
```
DEKHTIARJonathan, line 16: Can we make this "completely optional and default to `false` ? Or is it already ? 
Not have this mandatory.

At least in the `pyproject.toml`. In the JSON file it's machine generated we can make it mandatory.

DEKHTIARJonathan (approved):

PR #105, opened 2025-06-18
[Plugin] Dynamic Plugin API

DEKHTIARJonathan:
> [!CAUTION]  
> If you want to check this out, I advise using this branch: https://github.com/wheelnext/pep_xxx_wheel_variants/tree/interactive_plugins

This PR removes a significant portion of #99 in favor of a more flexible less "maintenance intensive" approach while allowing maximum freedom in plugin logic & sorting.

This PR introduces the concept of `PluginDynamicType`. The older `PluginType` becomes `PluginStaticType` (fully backward compatible).

- `PluginDynamicType` => takes the `VariantProperties` of the "candidate variants" in inputs
    -  They can use custom `data types`:
         - `float`
         - `SemVer`
         - `CalVer`
    - Any sorting logic they see fit to determine why `A :: B :: C` > `A :: B :: D`. 
    
- `PluginStaticType` => take no input and rely on pure `string matching` to operate

While `PluginDynamicType` is significantly more powerful, its implementation is significantly more delicate. And we should recommend `plugin authors` to use `PluginStaticType` interface as often as possible and rely only use `PluginDynamicType` when absolutely necessary.

Not only this PR enables to grasp fully the vast diversity of the variant ecosystem:
   - `float/int comparison` (e.g. minimum RAM requirement)
   - Version Schema (e.g. SemVer, CalVer, ARMVer ??)
   - etc.
   
It does it in a significantly more flexible, less arbitrary, more cleanly and more efficiently than PR #99.
The latter statement can be appreciated with the "fairly reduced scope" of the PR compared to #99.


DEKHTIARJonathan:
@mgorny if you check out https://github.com/wheelnext/pep_xxx_wheel_variants/tree/interactive_plugins - you can test completely the concept with `dummy-project` and `sandbox-project` in mockhouse.

I still need to fix some unittests. I'll do that later :)

variantlib/plugins/_subprocess.py
```diff
@@ -46,19 +57,30 @@ def load_plugins(plugin_apis: list[str]) -> Generator[PluginType]:
         else:
             plugin_instance = plugin_callable
 
-        required_attributes = PluginType.__abstractmethods__
+        if isinstance(plugin_instance, PluginStaticType):
```
mgorny, line 60: `isinstance()` won't work here since the protocol allows optional methods (`get_build_setup`). Also, I'm not sure if it's going to work with `ModuleType` and so on.

variantlib/plugins/_subprocess.py
```diff
@@ -71,41 +93,137 @@ def process_configs(
...
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "-p",
         "--plugin-api",
         action="append",
         help="Load specified plugin API",
         required=True,
     )
+
+    parser.add_argument(
```
mgorny, line 150: Just pass them via JSON. Then you won't need most of the boilerplate.

variantlib/plugins/_subprocess.py
```diff
@@ -71,41 +93,137 @@ def process_configs(
...
+@dataclass(frozen=True)
+class VariantProperty:
+    namespace: str
+    feature: str
+    value: str
+
+    @classmethod
+    def from_str(cls, input_str: str) -> Self:
+        # Try matching the input string with the regex pattern
+        match = VALIDATION_PROPERTY_REGEX.fullmatch(input_str.strip())
```
mgorny, line 124: The whole point of subprocess is to keep validation in the parent, and pass only validated data.

variantlib/protocols.py
```diff
@@ -84,3 +84,35 @@ def get_build_setup(
...
+        """Plugin namespace"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_all_features(self) -> list[VariantFeatureName]:
+        """Get all feature names for the plugin"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def filter_and_sort_properties(
```
mgorny, line 105: Just a random thought but I do wonder if it would be easier or harder to have plugins return sort keys (or `None` for unsupported variants) instead, and handle sorting and filtering on `variantlib` end. May or may not involve less duplication.

PR #106, opened 2025-06-23
Introduce support for dynamic config return values in plugins

mgorny:
Add support for the plugins dynamically adjusting the return values for `get_all_configs()` and `get_supported_configs()` based on input `known_properties` values.  The values are either provided based on the variants being validated (for `get_all_configs()`) or based on all properties found in available wheels (for
`get_supported_configs()`).  The dynamic API is entirely optional -- the plugin can simply ignore the additional argument and return a fixed list as usual.  It also permits for hybrid plugins that combine static and dynamic properties.


mgorny:
Now https://github.com/wheelnext/pep_xxx_wheel_variants/tree/dynamic-plugins-mgorny with plugins ported.

tests/artifacts/test-plugin-package/test_plugin_package.py
```diff
@@ -8,16 +15,19 @@ class FeatConfig:
 
 
 namespace = "installable_plugin"
+dynamic = False
 
 
-def get_all_configs() -> list[FeatConfig]:
+def get_all_configs(known_properties: Collection[VariantPropertyType] | None) -> list[FeatConfig]:
```
DEKHTIARJonathan, line 21: Let's change `Collection` to `frozenset`

tests/models/test_provider.py
```diff
@@ -125,7 +125,7 @@ def test_failing_regex_value() -> None:
     with pytest.raises(ValidationError, match="must match regex"):
         _ = VariantFeatureConfig(name="name", values=[""])
 
-    for c in "@#$%&*^()[]?!-{}[]\\/ ":
```
DEKHTIARJonathan, line 128: Why did you remove that ? Maybe remove `!` which is now allowed. And you can add `\t`

variantlib/constants.py
```diff
@@ -29,7 +29,7 @@
 
 VALIDATION_NAMESPACE_REGEX = re.compile(r"[a-z0-9_]+")
 VALIDATION_FEATURE_NAME_REGEX = re.compile(r"[a-z0-9_]+")
-VALIDATION_VALUE_REGEX = re.compile(r"[a-z0-9_.]+")
+VALIDATION_VALUE_REGEX = re.compile(r"\S+")
```
DEKHTIARJonathan, line 32:
```suggestion
VALIDATION_VALUE_REGEX = re.compile(r"[a-z0-9_.,!>~<=]+")
```

DEKHTIARJonathan (changes requested):
Really like the approve. Let's change the few details we discussed on Slack.
I still need to test and compare side by side with #105

PR #108, opened 2025-06-25
Support custom variant labels

mgorny:
Add support for a more generic concept of "variant labels" in filenames, to replace narrower "variant hashes". The label is still the hash by default, but the wheel builder can override it to any alphanumeric string up to 8 characters and underscores, providing a more readable description of the specific variant.

Variant label is the last component of the variant wheel filename, in place of the variant hash. It is also used as the key in `variants.json` files.

- `VariantDistInfo` now permits modifying `variant_label` property (formerly `variant_hash`) to change the label.
- `make-variant` command accepts a new `--variant-label=` option.
- `get_variant_hashes_by_priority()` is renamed to `get_variants_by_priority()`, and now returns variant labels rather than hashes.
- `make_variant_dist_info()` accepts an optional `variant_label` that can be used to override the label.


variantlib/constants.py
```diff
@@ -25,7 +25,7 @@
 VARIANTS_JSON_SCHEMA_URL = "https://variants-schema.wheelnext.dev/"
 VARIANTS_JSON_VARIANT_DATA_KEY: Literal["variants"] = "variants"
 
-VALIDATION_VARIANT_HASH_REGEX = re.compile(rf"[0-9a-f]{{{VARIANT_HASH_LEN}}}")
+VALIDATION_VARIANT_LABEL_REGEX = re.compile(r"[0-9a-z]{1,8}")
```
DEKHTIARJonathan, line 28: Please use `[0-9a-z_]{1,8}`

Let's allow `_` but not `.` or `-` => to avoid breaking `.split()` logic.

DEKHTIARJonathan:
Looking at the resulting filenames - it's really weird:

```
tests/artifacts/test-package/dist/test_package-0-py3-none-any-5d8be4b9.whl
tests/artifacts/test-package/dist/test_package-0-py3-none-any-60567bd9.whl
tests/artifacts/test-package/dist/test_package-0-py3-none-any-fbe82642.whl
tests/artifacts/test-package/dist/test_package-0-py3-none-any-bar.whl
tests/artifacts/test-package/dist/test_package-0-py3-none-any-foo.whl
```

I wonder if @konstin doesn't make a really good point about "separators".

**Maybe** one of these (or something else) is "safer":
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any#foo.whl`
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any-#foo.whl`
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any-!foo.whl`
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any+foo.whl`

The current "filename" allows **VERY** weird stuff:
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any-pypi.whl`
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any-python2.whl`
- `tests/artifacts/test-package/dist/test_package-0-py3-none-any-manylinux.whl`
- `...`

Now that these strings "(can) have a meaning" I find it very disturbing that the separator is a normal `-`.
It becomes pretty hard to distinguish in 1/4sec variant or not variant. 

The `hash` use to provide immediate visual feedback - immediately giving a visual cue "this is not a normal wheel".
I **feel** that the current "filename format" doesn't allow a strong enough visual difference to flag `-manylinux` is not platform tag, it's a variant name.

Sorry to not have made this comment earlier - it really just "stroke me now"

variantlib/constants.py
```diff
@@ -25,7 +25,7 @@
 VARIANTS_JSON_SCHEMA_URL = "https://variants-schema.wheelnext.dev/"
 VARIANTS_JSON_VARIANT_DATA_KEY: Literal["variants"] = "variants"
 
-VALIDATION_VARIANT_HASH_REGEX = re.compile(rf"[0-9a-f]{{{VARIANT_HASH_LEN}}}")
+VALIDATION_VARIANT_LABEL_REGEX = re.compile(r"[0-9a-z_]{1,8}")
```
DEKHTIARJonathan, line 28:
```suggestion
VALIDATION_VARIANT_LABEL_REGEX = re.compile(rf"[0-9a-z_]{1,{{VARIANT_HASH_LEN}}}")
```

This would be slightly better

DEKHTIARJonathan (commented):
Aside of my comment on `filename` - I really like this PR. Well made I like the approach and really doesn't change much of the internal mechanic.

Good job - let's wrap up the discussion the filename and we can merge this

tests/commands/test_make_variant.py
```diff
@@ -82,10 +86,13 @@ def test_make_variant(
             itertools.chain.from_iterable(["-p", vprop] for vprop in properties)
         )
 
+    if len(label) != 8:
+        cmd_args.append(f"--variant-label={label}")
```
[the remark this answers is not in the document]
mgorny: It's just a test case. This assumption makes it simpler.

mgorny:
I don't think "people using weird names" is a valid argument. Mind you, the following is also a valid wheel:

```
foo-1.2.3-py3.win64-none.linux-any.py36.whl
```

Going out of our way to prevent people from doing confusing things is not a good design goal. Changing the separator means that variant wheels will be wrongly taken by older `pip` versions which goes against our initial goals.

DEKHTIARJonathan:
> Changing the separator means that variant wheels will be wrongly taken by older pip versions which goes against our initial goals.

Would it ? It wouldn't fail the regex ? Because that's how I implemented Variants a while back. The first design used `-#abcd1234`

variantlib/constants.py
DEKHTIARJonathan: Fair enough - closing

mgorny:
Also note that Poetry's regexp is even terser — it's `[^-]+`, so `-` is really the only separator that works. We could add some extra character after that, but it feels like making the filename longer for a corner case of a corner case.

In the end, we are supposed to provide tools and guidance. We are not supposed to force people to use these tools "our way".

mgorny:
> * generation of the `variants.json` fails if the null variant has a label

That's going to be messy, since the tools aren't supposed to permit you to create such an artifact. I'd rather not add a backdoor to facilitate tests, or provide artifacts that can't be regenerated trivially.

mgorny:
Done for `make-variant` and `variants.json` parsing. The latter implies that all API should reject it now, as well as `generate-index-json`.

PR #110, opened 2025-08-08
Null variant labels

mgorny:
1. Add a `get_variant_label()` helper for convenience.
2. Use `null` label for null variants.
3. Change null variant hash to `sha256("")`.


DEKHTIARJonathan:
You can merge once you're happy with the changes - I'm happy with the implementation (aside of being more drastic in the approach)

tests/commands/test_make_variant.py
```diff
@@ -109,12 +110,17 @@ def test_make_variant(
...
+                "--variant-label=123456789",
+            ],
+            "Invalid variant label: '123456789' (must be up to 8 alphanumeric "
+            "characters)",
         ),
         (
-            ["--null-variant", "--variant-label=null"],
-            "error: --variant-label cannot be usedwith --null-variant",
+            ["--null-variant", "--variant-label=zuul"],
+            "Null variant must always use 'null' label",
```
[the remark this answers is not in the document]
mgorny: But it's precisely testing that an error is thrown?

variantlib/api.py
```diff
@@ -233,3 +223,39 @@ def get_variant_environment_dict(
...
+    """
+
+    if custom_label is None:
+        return (
+            NULL_VARIANT_LABEL
+            if variant_desc.is_null_variant()
+            else variant_desc.hexdigest
+        )
+
+    if variant_desc.is_null_variant():
```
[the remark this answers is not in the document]
mgorny: But then it would fail unless you specify `custom_label="null"` which is exactly what we don't want people to do?

variantlib/constants.py
```diff
@@ -5,7 +5,8 @@
 from typing import TypedDict
 
 VARIANT_HASH_LEN = 8
-NULL_VARIANT_HASH = "0" * VARIANT_HASH_LEN
+NULL_VARIANT_HASH = "e3b0c442"  # sha256sum("")
```
DEKHTIARJonathan, line 8: Is it really useful ? It can be felt as a magic value - and it's used only once inside `api.py`

variantlib/commands/make_variant.py
```diff
@@ -106,15 +106,6 @@ def make_variant(args: list[str]) -> None:
 
     parsed_args = parser.parse_args(args)
 
-    if parsed_args.variant_label is not None:
-        if parsed_args.null_variant:
-            parser.error("--variant-label cannot be usedwith --null-variant")
```
DEKHTIARJonathan, line 111: I would keep this - avoid silent "overwrite" and explicitly fail.
I may even keep the full block of code because the error is explicit - easy to read and parse

variantlib/constants.py
mgorny: Not really — I thought you wanted me to keep it.

variantlib/commands/make_variant.py
mgorny: It doesn't overwrite silently — it just fails with the exception coming from `variantlib.api`. My idea was to avoid code duplication.

PR #111, opened 2025-08-16
Support `none` namespace for non-plugin properties

mgorny:
Reserve a special namespace called `none` for the purpose of declaring variant properties that aren't handled by any plugin but the package itself. The valid features and their values, as well as their ordering, are fully specified in `pyproject.toml` then, for example:

```
[variant.default-priorities]
namespace = [..., "none"]
feature.none = ["blas", "build"]
property.none.blas = ["accelerate", "openblas", "mkl", "netlib"]
property.none.build = ["release", "debug"]
```

The following rules apply:

1. If the `none` namespace, the feature name and the value are present in `variant.default-priorities`, the property is considered both valid and supported. Otherwise, it's considered invalid.

2. `variant.providers.none` must never be specified (since it can't use a plugin).

3. Ordering is implied by the `variant.default-priorities` table, the same as it would sort regular properties.

For the implementation, I've chosen to:

1. Add a `VariantInfo.get_none_properties()` method that assembles the dictionary of valid/supported properties from the variant information.

2. Modify `BasePluginLoader` to account for `none` properties specially. This implies that we technically can skip plugins entirely when dealing with variants that use `none` properties only.

See https://github.com/wheelnext/pep_xxx_wheel_variants/issues/18

### DIFF
--- a/variantlib/constants.py
+++ b/variantlib/constants.py
@@ -8,6 +8,7 @@ VARIANT_HASH_LEN = 8
 NULL_VARIANT_LABEL = "null"
 CONFIG_FILENAME = "variants.toml"
 VARIANT_DIST_INFO_FILENAME = "variant.json"
+NAMESPACE_NONE = "none"
 
 # Common variant info keys (used in pyproject.toml and variants.json)
 VARIANT_INFO_DEFAULT_PRIO_KEY: Literal["default-priorities"] = "default-priorities"
--- a/variantlib/models/variant_info.py
+++ b/variantlib/models/variant_info.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING
 from typing import Any
 
 from packaging.requirements import Requirement
-
+from variantlib.constants import NAMESPACE_NONE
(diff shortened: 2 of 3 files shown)


mgorny:
For now, without tests, as I'd like to know your opinion first. I have mixed feelings about adding special casing like this, but I think it's useful and doesn't make things much more complex.

rgommers:
I was thinking about how to use this `none` namespace (before looking at this PR) here: https://github.com/wheelnext/pep_xxx_wheel_variants/issues/54#issuecomment-3199167808.

Looking at this example, my main questions are:
- How would you handle having a build-time plugin that deals with providing compiler flags? 
- That build-time plugin could also standardize the property name and its corresponding values. It feels like this is now completely ad-hoc and could differ across packages? (maybe `blas` isn't the best example here then?)
- Is there a concern about some values being only supported on specific platforms (e.g., `accelerate` is macOS-only)?

mgorny:
Closing in favor of #112

PR #115, opened 2025-09-08
Add a `has_fixed_supported_configs` plugin property

mgorny:
Add a `has_fixed_supported_configs` that indicates whether `get_supported_configs()` always returns a fixed list that does not depend on the system state.  Only plugins featuring this flag can be used with `plugin-use = "build"`.


konstin:
Make sense!

PR #121, opened 2025-10-03
Remove support for plugin instantiation

mgorny:
From now on, plugin classes are only allowed to use static or class methods.

Fixes https://github.com/wheelnext/pep_xxx_wheel_variants/issues/99


mgorny:
Added tests for protocol correctness, fixed missing `@abstractmethod` and switched to `@classmethod` whenever possible — if only to make documentation clearer.

PR #132, opened 2026-04-10
Fallback sorting on labels

mgorny:
Effort to add the fallback sort on variant label. After much testing, I've figured out that the only sane way to do this without having to change all the API is to add label to `VariantDescription`. For now, we're not verifying it yet, since there's still tons of call sites to update.


mgorny:
Okay, this is bigger than I initially planned but I think it's better to make it more future-proof.

The key changes are:
1. `VariantDescription` now carries the label in addition to variant properties. If it's not provided explicitly, we default to `self.hexdigest` or `null`.
2. The separate `label` argument to the API functions is now deprecated, but still supported — in which case it overrides the label found in `VariantDescription`. However, for future code we expect users to set `label` on `VariantDescription`.
3. Consequently, the `get_variant_label()` helper is now deprecated in favor of `VariantDescription.label`.
4. The sort algorithm is updated to take variant label as the final sort key, per the PEP. I've added extra tests for that.
5. I've updated all the tests to explicitly pass `label=` when constructing non-null `VariantDescription`. This is not technically necessary now; I've temporarily made that a requirement to make sure I didn't miss any call site. That said, I suppose the tests are more future-proof that way, in case we ever make labels required.
6. In data types that carry the label explicitly (dicts), we preserve the original format but aim to sync the label between the keys and `VariantDescription`s.

konsti-openai:
We need to remove duplicate filtering for this: https://github.com/wheelnext/variantlib/blob/bed4a86ace11ed21ee9bf103c508aef758f2c473/variantlib/resolver/lib.py#L90

variantlib/models/variant.py
```diff
@@ -179,6 +184,24 @@ def __post_init__(self) -> None:
             # Only "legal way" to modify a frozen dataclass attribute post init.
             object.__setattr__(self, "properties", sorted(self.properties))
 
+        # default the label to "null" or hexdigest
+        if not self.label:
+            with contextlib.suppress(AttributeError):
+                object.__setattr__(
```
konsti-openai, line 190: Why does this need `__setattr__`?

konsti-openai (approved):
LGTM apart from the duplicate filtering

variantlib/models/variant.py
mgorny: Because it's a frozen `dataclass`. Not that I like this design.

